PR #17, opened 2026-06-24
[codex] Support strided DS4 packed KV cache

voipmonitor:
## Summary

Allow SM120 compressed MLA to consume vLLM DeepSeek V4 packed KV cache views without materializing contiguous copies.

The change:
- accepts `[pages, page_bytes]` cache tensors whose last dimension is contiguous even when the page dimension has a larger packed block stride;
- derives the per-block byte stride from `cache.stride(0)` for both decode and prefill paths;
- maps non-contiguous packed views to the physical 1D byte span using `as_strided`, preserving storage offset and avoiding `.reshape(-1)` copies.

This is the B12X-side half of the DS4 packed-KV fix. The matching vLLM adapter PR exposes correctly offset page views for each DS4 cache layer.

## Root Cause

vLLM's DS4 packed-KV layout stores all per-layer cache payloads in a single packed per-block allocation. B12X compressed MLA previously assumed each cache page tensor was contiguous and had an implicit stride equal to the page payload size.

After DS4 KV packing, the layer payload is contiguous inside a page, but page-to-page stride is the packed block stride. Flattening a non-contiguous view with `.reshape(-1)` either materializes a copy or loses the raw address/stride relationship required by the CUDA kernels.

## Validation

Validated with current eldritch vLLM plus the matching vLLM adapter PR on DS4F TP2, MTP off, FP8 KV, B12X attention/MoE/linear, clean image:

`voipmonitor/vllm:eldritch-packedkv-vb923949-b12xf81d898-cu132-20260624`

Functional smoke:
- `/mnt/test.py --port 5627 -L -c 1200`: coherent Python output, CJK=0, generation-only ~`125 tok/s`.

Decode A/B, same image and B12X commit, same GPUs, comparing upstream packed KV vs temporary oldKV layout overlay:

| layout | C1 | C2 | C4 | C8 |
|---|---:|---:|---:|---:|
| packed KV | 119.2 | 190.1 | 305.4 | 448.8 |
| oldKV overlay | 119.8 | 190.7 | 308.0 | 450.3* |

`*` oldKV C8 was marked capacity-limited by the benchmark. Values are aggregate decode tok/s.

GPU 8-9-only PCIe monitor from the same decode run:

| layout | C1 rx/tx MB/s | C2 rx/tx MB/s | C4 rx/tx MB/s | C8 rx/tx MB/s |
|---|---:|---:|---:|---:|
| packed KV | 398/376 | 619/610 | 837/809 | 1284/1292 |
| oldKV overlay | 425/404 | 638/627 | 830/823 | 1089/1071* |

Prefill A/B:

| layout | 8k tok/s | 8k rx/tx MB/s | 64k tok/s | 64k rx/tx MB/s |
|---|---:|---:|---:|---:|
| packed KV | 8,632 | 15191/15982 | 8,151 | 14681/15420 |
| oldKV overlay | 8,748 | 13255/13103 | 8,148 | 15089/14940 |

Interpretation: packed-KV support restores B12X to the oldKV throughput range without reverting vLLM's packed allocation. PCIe numbers are coarse `nvidia-smi dmon` diagnostics and were collected in a helper container that could only see GPUs 8 and 9.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cache handling for attention decoding and prefill so packed or non-contiguous cache layouts are processed correctly.
  * Added stricter validation for cache layouts and stride sizes, reducing launch-time errors with unsupported tensor shapes.
  * Fixed paged top-k scoring to automatically fall back to a compatible path when the cache layout does not match the packed-contiguous assumption.

### DIFF
--- a/b12x/attention/indexer/paged.py
+++ b/b12x/attention/indexer/paged.py
@@ -681,6 +681,16 @@ def index_topk_fp8(
             "paged index supertile top-k scratch page-table capacity is too small: "
             f"need={page_table_width}, have={getattr(scratch, 'max_page_table_width', None)}"
         )
+    if (
+        route == "packed_contiguous"
+        and int(index_k_cache.stride(0)) != int(_PAGED_INDEX_CACHE_ROW_BYTES)
+    ):
+        # DS4 packed KV stores each layer as a strided view into a larger
+        # per-block allocation. The packed-contiguous prefill scorer first gathers
+        # a shared supertile into contiguous scratch; that path is only valid for
+        # physically contiguous page rows. The tiled scorer consumes the runtime
+        # cache stride and keeps the vLLM packed-KV layout intact.
+        route = "paged_tiled"
     use_shared_prefill_scorer = route == "packed_contiguous"
     topk_block_k = (
         int(getattr(binding, "prefill_block_k", 0) or getattr(scratch, "prefill_block_k", 0))
--- a/b12x/attention/mla/compressed_api.py
+++ b/b12x/attention/mla/compressed_api.py
@@ -421,8 +421,11 @@ def _compressed_mla_cache_byte_view(cache: torch.Tensor, *, name: str) -> torch.
         )
     if byte_cache.ndim != 2:
         raise ValueError(f"{name} must have shape [pages, page_bytes], got {tuple(cache.shape)}")
-    if not byte_cache.is_contiguous():
-        raise ValueError(f"{name} must be contiguous for compressed MLA")
+    if int(byte_cache.stride(1)) != 1:
+        raise ValueError(
+            f"{name} page payload must be contiguous in the last dimension, "
+            f"got stride {tuple(byte_cache.stride())}"
+        )
     return byte_cache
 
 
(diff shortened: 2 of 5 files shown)


voipmonitor:
Added `46c7fc4` to cover the DS4 packed-KV sparse-indexer case.

Root cause: after vLLM switched DS4 to packed KV, each layer's index cache is a strided view into a larger per-block allocation. The B12X `packed_contiguous` indexer route first gathers a shared supertile and assumes physically contiguous page rows, so it is not a valid route for that layout. The patch keeps packed KV enabled and falls back to the tiled paged scorer only when `index_k_cache.stride(0)` is not the physical contiguous page-row size. Contiguous layouts still keep the existing route.

Validation on the packed-KV image with this PR overlay, TP2, GPUs 8,9, port 5630:

- `CUDA_LAUNCH_BLOCKING=1`, `max_num_seqs=128`, graph 4, C128/ctx0/max256: passed, aggregate decode 381.1 tok/s, no CUDA illegal access.
- Full mode, graph 256, C128/ctx0/max2048: passed, aggregate decode 1886.1 tok/s, max_running_reqs=128, no `EngineDead`/traceback.

This is not reverting packed KV; it only prevents the indexer from selecting a route whose memory-layout contract does not match strided packed KV.

PR #21, opened 2026-07-02
perf(gemm): 8-35x faster dense MXFP8 input quantizer + SF load reductions

voipmonitor:
Two changes to the dense MXFP8 linear path, both bit-exact, measured on DSV4-Flash TP2 shapes on RTX PRO 6000 Blackwell (sm_120):

## 1. Tile the dense MXFP8 input quantizer (`a67e5bd`)

`_quantize_dense_tk_to_tk_kernel` launched one CTA per (token, 32-elem group) — ~1.05M CTAs at M=8192/K=4096 — and cost 229 ms per 4-step DS4 TP2 prefill window vs 14 ms for DeepGEMM's per-token-group quant. Retiled to `BLOCK_M=32 x GROUPS=16` per program with masked edges and vectorized stores (incl. the swizzled `scale_mma` layout):

| K (M=8192) | before | after | speedup |
|---|---:|---:|---:|
| 1024 | 122.5 us | 11.9 us | 10.3x |
| 4096 | 483.6 us | 13.7 us | 35.2x |
| 8192 | 965.5 us | 118.1 us | 8.2x |

Bit-exact vs the old kernel and an independent torch reference (values, `scale_rows`, swizzled `scale_mma`), incl. M=1/4/5/33/100/8191 and K=1152.

**E2E: DS4-Flash TP2 A8 full-B12X standalone prefill 12,468/11,971/11,133 → 13,441/12,940/11,948 tok/s @ 8k/64k/128k** (Lucifer CUTLASS parity at 8k, +2.5%/+2.0% above it at 64k/128k). Decode cc1 unchanged (ITL 7.72 ms). 30k-ctx coherence clean.

## 2. Stage-bulk SF copies + k-replicated SFB reuse (`49b453c`)

NCU showed the mainloop issuing 37.7M scalar `LDS` (per-byte SF loads) vs DeepGEMM's 1.3M for the same GEMM. Two bit-exact reductions:
- SF register fragments for the whole acquired pipeline stage load in one bulk copy instead of per-`k_block` slices.
- `sfb_k_replicated` flag (passed by `block_fp8_linear`): 128x128-block weight scales are expanded to per-32 replicas, so the four SFB bytes per 128-wide k tile are identical by construction — load one per stage and feed every k block from it. Guarded off under `swap_ab`. LDS drops to 9.4M.

Honest note: wall-clock is **unchanged** by (2) — SM throughput stays ~73%, which falsifies instruction count as the large-M limiter. Full NCU analysis (speed-of-light, opcode mix, stall profile), the falsified-lever list (tile sweep, tile_k, prefetch, atom_shape, region-illegal SF pre-binding), and a concrete mainloop rewrite spec for reaching DeepGEMM at M>=4096 natively are written up here:

https://github.com/voipmonitor/rtx6kpro/blob/master/optimization/b12x-dense-fp8-gemm-vs-deepgemm.md

## Validation

- `tests/test_gemm_block_fp8_linear.py`, `tests/test_fp8_quant_deepgemm_parity.py`, `tests/test_block_fp8_linear_scratch_bindings.py`, `tests/test_dense_gemm_expected_m.py`: **5 passed, 1 failed** — `test_block_fp8_linear_small_live_m_reuses_prefill_dense_kernel` fails identically on stock `77bd50e` in the same container (pre-existing in this environment, not introduced here).
- Byte-identical GEMM outputs with the flag on vs off on packer-built weights (all four DSV4 TP2 dense shapes, M=8192).
- Served E2E on DSV4-Flash TP2 A8 (B12X_MLA_SPARSE + b12x MoE + b12x linear) for prefill/decode/coherence benches.

### DIFF
--- a/b12x/gemm/block_fp8_linear.py
+++ b/b12x/gemm/block_fp8_linear.py
@@ -43,6 +43,7 @@ _B12X_TIMING_THRESHOLD_MS = float(
 _SCRATCH_ALIGN_BYTES = 1024
 
 
+
 @dataclass(frozen=True)
 class BlockFP8LinearWeight:
     weight: MXFP8Rows
@@ -664,6 +665,9 @@ def _block_fp8_linear_mxfp8_fused_op(
         c_dtype=_c_dtype_name(source_2d.dtype),
         sf_vec_size=MXFP8_SCALE_VEC_SIZE,
         expected_m=expected_m,
+        # Weight scales come from 128x128 blocks expanded to per-32 rows, so
+        # the four SFB bytes per 128-wide k tile are identical by construction.
+        sfb_k_replicated=True,
         stream=stream_int,
     )[:, :, 0]
 
@@ -792,6 +796,7 @@ def block_fp8_linear_mxfp8(
         sf_vec_size=MXFP8_SCALE_VEC_SIZE,
         out=output_storage,
         expected_m=expected_m,
+        sfb_k_replicated=True,
         stream=stream,
     )[:, :, 0]
     t_gemm = time.perf_counter() if _B12X_TIMING else 0.0
--- a/b12x/gemm/block_fp8_linear.py
+++ b/b12x/gemm/block_fp8_linear.py
@@ -165,6 +165,7 @@ def _quantize_dense_tk_to_tk_kernel(
     scale_rows,
     scale_mma,
     tokens,
+    num_groups_k,
     source_stride_t,
     source_stride_k,
     values_stride_t,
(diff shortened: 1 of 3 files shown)


voipmonitor:
Added `tests/test_w4a8_rp_inverse_mapping.py` (123a16f): bit-exact inverse formulas for the W4A8-MX N256/K128 in-place repack (weights + e8m0 sfb grids, both w13-rotated and w2 orientations), verified against the real prep kernels. This is the groundwork for a tiny-decode band that reads the repacked buffers directly — context: at DSV4-Flash TP2 M=1 decode, the w4a8_mx dynamic path runs 34.8 µs/layer vs the w4a16 fused kernel's 22.5 µs (× 61 MoE layers = the whole A8-vs-A16 decode gap, ~130 vs ~140 tok/s E2E).

One design constraint worth knowing before building it: the rp 16B-contiguous unit spans logical rows {r, r+8, r+16, r+24} at one k-window, so a row-per-warp dataflow (like direct-micro's FC1 v4 loads) drops to ~12.5% sector efficiency under this layout — the lane map needs to cover four 8-apart rows per instruction. Full measurements and the falsified-knob list (MAC ladder sweep, tile_n≠128, atom (1,8,1) FPE) are in https://github.com/voipmonitor/rtx6kpro/blob/master/optimization/b12x-dense-fp8-gemm-vs-deepgemm.md §6.8.

PR #25, opened 2026-07-05
Add docker/ source-build workflow (build b12x from checkout into the vLLM images)

jarrelscy:
Adds `docker/Dockerfile` + `docker/README.md`: overlay the current source checkout onto a `voipmonitor/vllm` base image with `pip install --no-deps --force-reinstall`, so kernel changes can be tested end-to-end (vLLM serve, CUDA graphs, TP, or the pytest suite) without rebuilding vLLM. Docs cover the JIT model (no AOT build; `CUTE_DSL_ARCH=sm_120a` at engine start), running tests in the image, and choosing a different base via `--build-arg BASE=`.

Verified: builds against `eldritch-enlightenment-v45c1582` and the source-installed b12x imports cleanly; this is the workflow I used to validate #24 on 4x RTX PRO 6000.

## Summary by CodeRabbit

* **New Features**
  * Added Docker support for building a development/serving image from the current source.
  * The image can reuse a configurable base image and preserves the existing runtime environment while updating the app package.
  * Sets GPU-targeted runtime settings for kernel compilation at startup.

* **Documentation**
  * Added setup and usage instructions for building, testing, and running the containerized environment.
  * Included guidance for overriding the base image during builds.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+.git
+*.ncu-rep
+__pycache__
+.pytest_cache
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,23 @@
+# Build a vLLM serving image with b12x installed from this source checkout.
+#
+# The published voipmonitor/vllm images ship a pinned b12x; this overlays the
+# repo's working tree on top so kernel changes can be tested end-to-end
+# without rebuilding vLLM. b12x kernels are pure-Python CuTe DSL — there is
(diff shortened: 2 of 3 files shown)


lukealonso:
Thanks for the other PR, but I don't think I want to merge this one since it inverts the natural dependency order (this belongs in github.com/local-inference-lab/vllm most likely)

jarrelscy:
Makes sense!

PR #26, opened 2026-07-07
moe(tiny_decode): support odd FC2 K-tile counts (n % 128 shards)

voipmonitor:
## Problem

`tiny_decode` (the M≤4 W4A8-MX decode kernel) requires `n % 256 == 0` and `(n // 128) % 2 == 0`, so padded GLM-5.2 TP6 expert shards (`moe_intermediate 2048/6` → 384 = 3×128, produced by the vLLM-side QMMA padding fix in [vllm#80](https://github.com/local-inference-lab/vllm/pull/80)) never take the tiny path — M≤4 A8 decode falls back to the dynamic QMMA kernels, costing ~13% decode vs A16 at TP6.

## Fix

Structurally the rp tile grid only needs whole 256-row tiles per GEMM **N** dim (FC1 tiles `2n` rows, so `n % 128` suffices; FC2's N dim is `k`). The only real 256 dependency was FC2's fixed two-K-tiles-per-task grouping (`_FC2_KT_PER_TASK = 2`).

- Make the FC2 K-tiles-per-task a `configure()`-time value: **2 when the tile count is even** (unchanged behavior, unchanged compiled kernels for every existing shape — the value is part of the cfg/cache key), **1 when odd**. FC2 partials are scatter-added, so the finer task split cannot change results.
- Relax `configure()` validation and `_tiny_decode_supports` to `n % 128 == 0`.

## Tests

Parametrized `test_w4a8_mx_small_band_matches_fp32_oracle` over `n ∈ {1024, 384}`: the 384 cases assert the planner selects the micro path and match the fp32 reference — 6/6 pass (m ∈ {1,2,4}).

## E2E

GLM-5.2-BF16-AMDMXFP4experts, TP6/DCP1, force-A8 + online MXFP8 dense overlay, shards padded 352→384:

| | A8 before | **A8 + this fix** | A16 |
|---|---:|---:|---:|
| decode c0 | 72.2 | **83.0** | 79.2 |
| decode c3000 | 71.2 | **81.4** | 77.9 |
| prefill 8k / 64k | 4,770 / 5,225 | 4,864 / 5,219 (unchanged) | 4,643 / 4,816 |

A8 goes from the slowest to the fastest TP6 mode on both axes; outputs coherent (0 CJK).

## Summary by CodeRabbit

* **New Features**
  * Expanded tiny-decode and MoE kernel eligibility by relaxing intermediate/FC geometry constraints and FC2 tiling rules.
  * Improved support for more intermediate-size shapes in small-band execution.

* **Bug Fixes**
  * Relaxed overly strict divisibility requirements that could incorrectly reject valid configurations.
  * Updated FC2 phase scheduling so task work partitioning adapts to the computed configuration.

* **Tests**
  * Enhanced oracle-matching coverage by running reference comparisons across multiple `n` values (in addition to existing `m` coverage).

### DIFF
--- a/b12x/integration/tp_moe.py
+++ b/b12x/integration/tp_moe.py
@@ -8015,13 +8015,14 @@ def _tiny_decode_enabled() -> bool:
 
 
 def _tiny_decode_supports(*, num_tokens: int, k: int, n: int, activation: str) -> bool:
+    # n needs whole 128-column rp tiles only; odd tile counts (e.g. n=384
+    # from GLM 2048/TP6 padded shards) run FC2 with one K tile per task.
     return (
         1 <= num_tokens <= 4
         and activation == "silu"
         and k % 256 == 0
-        and n % 256 == 0
+        and n % 128 == 0
         and (k // 128) % 4 == 0
-        and (n // 128) % 2 == 0
     )
 
 
--- a/b12x/moe/fused/tiny_decode.py
+++ b/b12x/moe/fused/tiny_decode.py
@@ -84,15 +84,24 @@ class MoETinyDecodeKernelBackend:
         device: torch.device | None = None,
     ) -> None:
         del device
-        if k % 256 != 0 or n % 256 != 0:
-            raise ValueError("tiny_decode requires k % 256 == 0 and n % 256 == 0")
+        # The rp tile grid needs whole 256-row tiles on each GEMM's N dim:
+        # FC1 tiles 2n rows (so n % 128 suffices) and FC2 tiles k rows.
+        if k % 256 != 0 or (2 * n) % 256 != 0:
+            raise ValueError("tiny_decode requires k % 256 == 0 and n % 128 == 0")
         if m < 1 or m > 4:
             raise ValueError("tiny_decode supports 1 <= m <= 4")
         rt = m * num_topk
         kt13 = k // 128
         kt2 = n // 128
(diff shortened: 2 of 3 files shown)


voipmonitor:
Measured follow-up on the W4A16 small-M path (asked to close the tiny-vs-TC gap): isolated graph-replay A/B on GLM-5.2 shard shapes (E=256, K=6144, N=384, topk=8, M=1, same e8m0 weights):

| M=1 MoE call | graph replay |
|---|---:|
| tiny_decode (rp layout, bf16 acts) | **16.2 µs** |
| W4A16 TC-decode (packed layout) | **32.8 µs** |

outputs cos 0.99996 — tiny already computes the w4a16 recipe (bf16 activations), so at M≤4 the two modes are numerically interchangeable; TC-decode has a ~2× per-call reserve at M=1 (E2E it shows as ~5% decode because the MoE is only part of the step).

I prototyped serving the W4A16 M≤4 band with tiny via an auxiliary rp-repacked copy (env-gated, oracle tests passed) but killed it at E2E: the aux copy is per-MoE-layer, i.e. a full second expert allocation (+46 GB/rank on GLM TP8) — weight duplication cannot close this. The only real equalization is teaching tiny_decode (or a sibling) the MMA_PACKED layout so W4A16 mode can reuse it without a second copy — leaving that call to you; the isolated numbers above say what it's worth. Until then, the practical guidance stands: on e8m0 checkpoints the fast A16-quality M≤4 decode is the w4a8_mx mode itself (tiny, bf16 acts).

PR #28, opened 2026-07-09
Handle plain CuTe compile callable with DSL options

voipmonitor:
## Summary
- keep the existing DSL compile options path for W4A8 dynamic MoE
- handle runtimes where cutlass.cute.compile is exposed as a plain function instead of a CompileCallable instance
- instantiate CompileCallable(dsl_compile_options) explicitly in that case, rather than crashing or dropping the options

## Root Cause
The GLM 5.2 v15 FF image hit TypeError: function object is not subscriptable in b12x/cute/compiler.py when W4A8 dynamic MoE requested OptLevel(2). In the vLLM worker runtime, cutlass.cute.compile can be a plain function, so cute.compile[dsl_compile_options] is not always valid.

## Validation
- runtime overlay benchmark on 8x RTX Pro 6000 Blackwell for /root/models/GLM-5.2-BF16-AMDMXFP4experts, TP8/DCP1/MTP0/A8 force
- before patch: first request killed EngineCore with TypeError at cute.compile[dsl_compile_options]
- after patch: mxfp4-a8-orig completed, decode cc1 88.612 tok/s, prefill 8k 6198 tok/s, prefill 64k 6365 tok/s, KV cache 625344 tokens

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compilation compatibility when DSL options are provided.
  * DSL options are now applied correctly whether the compiler is exposed as a configurable callable or a standard function.

### DIFF
--- a/b12x/cute/compiler.py
+++ b/b12x/cute/compiler.py
@@ -1994,7 +1994,17 @@ def compile(
         # Subscript-style DSL compile options (e.g. OptLevel(2): ptxas -O3's
         # scheduler register-starves some scalar-heavy kernels; see the w4a8
         # dynamic MoE recipe).
-        compile_callable = cute.compile[dsl_compile_options]
+        if hasattr(compile_callable, "__getitem__"):
+            compile_callable = compile_callable[dsl_compile_options]
+        else:
+            # Some embedded runtimes expose cutlass.cute.compile as a plain
+            # function instead of the CompileCallable instance installed by the
+            # top-level module import.  Recreate the callable explicitly so DSL
+            # options still take effect instead of crashing or silently falling
+            # back to the default compiler options.
+            from cutlass.base_dsl.compiler import CompileCallable
+
+            compile_callable = CompileCallable(dsl_compile_options)
         kwargs = dict(kwargs)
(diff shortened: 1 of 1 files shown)


voipmonitor:
Revalidated together with #32 on current B12X master 115926e. Both cherry-pick cleanly; tests/test_cutlass_runtime_patches.py plus tests/test_w4a8_mx_tp_moe.py completed with 84 passed, 1 skipped on SM120. No compatibility patch was required.

voipmonitor:
Canonical-stack validation update: current `master` (`115926e`) plus #28, #32, #34, and #35 merges without conflicts. The combined SM120 suite passed: 120 passed, 1 skipped (CuTe runtime patching, W4A8 auxiliary-stream launch, W4A16/NF3 launch geometry, and NVFP4 MLA KV cache/API coverage). `git diff --check` is clean.

voipmonitor:
Final canonical-stack validation against `master@115926eb`: merged #28 -> #32 -> #34 -> #35 without conflicts, then ran the combined production-path suite: **122 passed, 1 skipped**. `git diff --check` and Ruff diagnostics on the changed files are clean. No runtime overlay is required.

PR #30, opened 2026-07-12
moe(w4a16): NF3 3-bit weight layout — mixed-precision MoE serving (NVFP4 + NF3)

MadeBy561:
## What

A new `weight_layout="nf3_2p1"` (+ `scale_format="e4m3_k32"`) for the fused W4A16 MoE path: a 3-bit normal-float weight format that serves alongside packed NVFP4 through the same kernel machinery. Two preplanned launches per MoE layer (one per precision tier) share one route geometry and one `topk_sum`, both CUDA-graph-safe.

Motivation: instead of pruning experts to fit big MoE models, keep **all** experts and quantize the low-saliency tail to 3 bits. GLM-5.2 753B with all 256 experts/layer fits and serves on 4× 96 GB SM120 this way (top-64 experts NVFP4, 192-expert tail NF3), and has been our production configuration since 2026-07-05.

Four files, kernel-side only: `b12x/moe/fused/w4a16/{kernel,prepare,host}.py` + `b12x/cute/fp4.py`. The serving integration (runtime graft, checkpoint build/GPTQ pipeline, tests, proven configs) lives out-of-tree: [MadeBy561/b12x@nf3-hybrid-full](https://github.com/MadeBy561/b12x/tree/nf3-hybrid-full/nf3-hybrid), checkpoint at [madeby561/GLM-5.2-NVFP4-NF3-Hybrid](https://huggingface.co/madeby561/GLM-5.2-NVFP4-NF3-Hybrid).

## Kernel design

- **Staging**: NF3 packs 32 codes into a 12-byte (lo0, lo1, hi) triple; B-tile staging is byte-parameterized flat copy (12 B/unit vs 16 B int4), same cp.async machinery.
- **Fragment reads**: three scalar u32 loads over stride-3 words — bank-conflict-free (gcd(3, 32) = 1).
- **Dequant**: `prmt`-LUT over codebook pools; the 8-level codebook enters as a kernel parameter, not a constant — learned per-tensor codebooks are free later.
- **Scales**: `e4m3_k32` group-32 decode on the existing scale pipeline (2^-116 convention; compensation folded into the f32 global scale — zero new kernel scale math).
- **`tc_zero_output` const_expr flag**: lets a second TC-decode launch accumulate into a shared output without re-zeroing (the two-tier decode path).
- **Planner accuracy** (second commit): `weight_layout`-aware register table (NF3 measures 238 regs at the production tile config on SM120 ncu output vs the packed table's 255) and B-stage footprint (12 B/32-code = ×3/4). Verified inert at production tiles — planning truth for future tile/occupancy decisions on NF3 shapes.

## What is not touched

`packed` / `modelopt` planning and codegen are byte-identical — every new parameter defaults to prior behavior.

## Receipts (GLM-5.2 753B all-256-expert hybrid, 4× RTX PRO 6000 Max-Q, TP4/DCP4)

- **GPQA-Diamond 176/198 (88.89 %)** — statistically tied with full precision (89.52 % FP8 / 89.39 % all-NVFP4 at 435 GB, same harness).
- KLD vs the official-FP8 reference (fixed 2,047-position corpus, identical rig for every number): 0.126 per-token mean (same-rig anchors: 435 GB all-NVFP4 0.060, 340 GB UD-IQ4_XS 0.087).
- Kernel fidelity vs a Triton oracle: argmax 100 %, logit MAE 0.019 (same-server A/B).
- ncu at production tiles: zero register spills.
- Serving: pool 422,656 tokens @ 262k max-len, prefill 1.65–1.72k tok/s, decode 92–104 tok/s single-stream (MTP5).

## Testing

Pack/unpack, scale-convention, and kernel-probe tests live with the out-of-tree harness (GPU tests). The suite re-run against this exact branch is queued for our next free serving window — results will be posted in this thread before we ask for merge. Production serving on the rebased kernel code is continuous.

### DIFF
--- a/b12x/cute/fp4.py
+++ b/b12x/cute/fp4.py
@@ -5522,3 +5522,129 @@ def compute_y_and_max_abs_f32(
         max_abs = fmax_f32(max_abs, fabs_f32(y_f32[i]))
 
     return y_f32, max_abs
+
+
+# ---------------------------------------------------------------------------
+# NF3 (3-bit NormalFloat codebook) dequant for the W4A16 "nf3_2p1" layout.
+#
+# Input per call: 8 codes in 2+1 split-plane form — lo16 (8 x 2-bit fields,
+# code j bits 1:0 at bits 2j+1:2j) and hi8 (8 x 1-bit, code j bit 2 at bit j)
+# — plus the 8-entry bf16 codebook as 4 u32 pools (lo/hi bytes of entries
+# 0..3 and 4..7).  Output: 4 packed bf16x2 fragments in the SAME element
+# order as packed_dequant_e2m1x4 pairs: o0=(v0,v1) o1=(v2,v3) o2=(v4,v5)
+# o3=(v6,v7).
+#
+# Selector build is the carry-safe shift ladder (NOT a multiply spread — the
+# 0x249 multiply trick carries between lanes).  prmt pool selectors are the
+# 3-bit codes themselves; bit 3 of every selector nibble stays 0 so prmt
+# byte-select mode is used, never sign-replicate.
+# ---------------------------------------------------------------------------
+@dsl_user_op
+def packed_dequant_nf3x8_to_bfloat2x4(
+    lo16: Uint32,
+    hi8: Uint32,
+    cb_lo01: Uint32,
+    cb_lo23: Uint32,
+    cb_hi01: Uint32,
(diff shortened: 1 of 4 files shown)


MadeBy561:
@coderabbitai full review

MadeBy561:
Withdrawing to rework — will resubmit.

PR #31, opened 2026-07-12
moe(w4a16): nf3_2p1 — 3-bit normal-float weight layout

MadeBy561:
## What

`weight_layout="nf3_2p1"` + `scale_format="e4m3_k32"`: a 3-bit normal-float weight format for the fused W4A16 MoE path, servable alongside packed NVFP4 — two preplanned launches per MoE layer share one route geometry and one `topk_sum`, both CUDA-graph-safe. Purpose: keep **all** experts and quantize the low-saliency tail to 3 bits instead of pruning. GLM-5.2 753B with all 256 experts/layer serves on 4× 96 GB SM120 this way (production since 2026-07-05).

## Kernel notes

- 32 codes pack into one 12-byte (lo0, lo1, hi) triple; B staging is the existing cp.async machinery with byte-parameterized units (12 B vs 16 B int4).
- Fragment reads are three scalar u32 loads over stride-3 words — bank-conflict-free (gcd(3, 32) = 1).
- Dequant is a `prmt` LUT over codebook pools; the 8-level codebook is a kernel **parameter**, so learned per-tensor codebooks need no kernel change.
- `e4m3_k32` scales ride the existing scale pipeline (2^-116 convention; compensation folds into the f32 global scale — no new kernel scale math).
- A `tc_zero_output` const_expr flag lets a second TC-decode launch accumulate into a shared output without re-zeroing.
- Second commit: layout-aware planner accounting — NF3 measures 238 registers at the production tile config (SM120 ncu) vs the packed table's 255, and its B-stage footprint is ×3/4. Verified inert at production tiles; planning truth for future NF3 tile decisions.

**No new environment variables.** `packed`/`modelopt` planning and codegen are byte-identical — every new parameter defaults to prior behavior.

## Testing

`tests/test_w4a16_nf3.py` (this PR) packs with the production packer, compiles the `nf3_2p1` kernel, and matches a pure-torch dequant reference on the TC-decode (m=1, 5) and route-packed (m=33) paths — 3/3 green on RTX PRO 6000 (SM120). The existing w4a16 suites pass unchanged on this branch (route-pack, packed-format, reference, execution-model, e2e — 202 passed; one pre-existing flashinfer-prep failure reproduces identically on pristine master in the same environment). Production serving of this kernel code has been continuous since 2026-07-05.

## Quality receipts (GLM-5.2 753B all-256-expert hybrid, 4× RTX PRO 6000 Max-Q, TP4/DCP4)

- **GPQA-Diamond 176/198 (88.89 %)** — statistically tied with full precision (89.52 % FP8 / 89.39 % all-NVFP4 at 435 GB, same harness).
- KLD vs a full-precision BF16 reference (fixed 2,047-position corpus, identical rig and reference logits for every number): 0.126 per-token mean (same-rig anchors: 435 GB all-NVFP4 0.060, 340 GB UD-IQ4_XS 0.087).
- Kernel fidelity vs a Triton oracle: argmax 100 %, logit MAE 0.019 (same-server A/B).
- Serving: pool 422,656 tokens @ 262k max-len, prefill 1.65–1.72k tok/s, decode 92–104 tok/s single-stream (MTP5).

Serving integration (a runtime graft — no vLLM sources modified), checkpoint pipeline, and GPU tests are maintained out-of-tree; checkpoint and serving docs: [madeby561/GLM-5.2-MXFP8-NVFP4-NF3-Hybrid](https://huggingface.co/madeby561/GLM-5.2-MXFP8-NVFP4-NF3-Hybrid).

---

Co-authored with Claude (Fable 5).

### DIFF
--- a/b12x/cute/fp4.py
+++ b/b12x/cute/fp4.py
@@ -5522,3 +5522,129 @@ def compute_y_and_max_abs_f32(
         max_abs = fmax_f32(max_abs, fabs_f32(y_f32[i]))
 
     return y_f32, max_abs
+
+
+# ---------------------------------------------------------------------------
+# NF3 (3-bit NormalFloat codebook) dequant for the W4A16 "nf3_2p1" layout.
+#
+# Input per call: 8 codes in 2+1 split-plane form — lo16 (8 x 2-bit fields,
+# code j bits 1:0 at bits 2j+1:2j) and hi8 (8 x 1-bit, code j bit 2 at bit j)
+# — plus the 8-entry bf16 codebook as 4 u32 pools (lo/hi bytes of entries
+# 0..3 and 4..7).  Output: 4 packed bf16x2 fragments in the SAME element
+# order as packed_dequant_e2m1x4 pairs: o0=(v0,v1) o1=(v2,v3) o2=(v4,v5)
+# o3=(v6,v7).
+#
+# Selector build is the carry-safe shift ladder (NOT a multiply spread — the
+# 0x249 multiply trick carries between lanes).  prmt pool selectors are the
+# 3-bit codes themselves; bit 3 of every selector nibble stays 0 so prmt
+# byte-select mode is used, never sign-replicate.
+# ---------------------------------------------------------------------------
+@dsl_user_op
+def packed_dequant_nf3x8_to_bfloat2x4(
+    lo16: Uint32,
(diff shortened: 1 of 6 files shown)


lukealonso:
Looks fine, will check for any non-NF3 perf regressions and then merge.

PR #32, opened 2026-07-14
Fix W4A8 resident-grid launch beside auxiliary streams

voipmonitor:
## Root cause

The fused dynamic W4A8 kernel crosses several software resident-grid barriers, but it was launched with an ordinary CUDA kernel launch. At the M=9 boundary the SM120 path selects tile M=16 and a two-CTA-per-SM grid (376 CTAs on RTX 6000 Pro Blackwell). When shared-expert GEMM is active on the vLLM auxiliary stream, only part of that grid can become resident. Those CTAs spin in the grid barrier while unscheduled CTAs cannot arrive, eventually producing an illegal memory access / Xid 31 during FULL CUDA graph capture.

`CUDA_LAUNCH_BLOCKING=1` and disabling the shared-expert stream only masked the race; neither is part of this fix.

## Fix

- launch the dynamic kernel cooperatively so its all-CTA residency assumption is guaranteed by CUDA
- keep the CTA-leader predicate in CuTe DSL IR instead of converting a dynamic device boolean through a Python conditional
- add an M=9 CUDA-graph regression test with concurrent BF16 GEMM work on an auxiliary stream

## Validation

- targeted graph replay test: passed
- new M=9 auxiliary-stream contention test: passed
- exact GLM 5.2 TP6/DCP3/A8/graph64 startup: all PIECEWISE and FULL captures passed, including the formerly deterministic M=9 failure
- 30k-token prefill plus sustained decode replay: passed
- no new NVIDIA Xid after the corrected run

The overlap, MoE backend, quantization mode, grid size, and vLLM stream configuration remain enabled and unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fused MoE kernel reliability during CUDA Graph replay when concurrent GPU work runs on auxiliary streams.
  * Prevented potential synchronization deadlocks at grid-phase boundaries.

* **Tests**
  * Added regression coverage for w4a8 mixed-precision MoE graph replay with concurrent matrix operations.
  * Verified replay results remain finite and non-zero.

PR #34, opened 2026-07-17
moe(w4a16): preserve preplanned tile geometry across launch op

voipmonitor:
## Problem

`run_w4a16_moe()` receives a precompiled launch whose weight packing may be tile-specific, especially for `nf3_2p1`. The Torch custom-op boundary cannot carry that compiled object. Reconstructing the launch with automatic tile selection can choose different FC1/FC2 tiles for the runtime M shape, silently selecting a different cache entry than the one used to pack the weights.

## Fix

Thread the four selected tile dimensions through the custom-op and fake-op schemas and pass them back as `force_tile_config` when resolving the compiled launch. This preserves the exact preplanned geometry for normal and calibrated launches.

Also remove a stale unused `math` import already exposed by current-master lint.

## Validation

- Rebased directly on B12X master `115926e`
- Production-shape GLM-5.2 TP4 NF3 TC-decode test with tier-masked routes: passed on SM120
- W4A16 custom-op fake-dispatch schema test: passed
- `ruff check` on all changed files
- `ruff format --check` on changed tests
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fused MoE execution by carrying the selected kernel tile configuration through the custom-op launch path, including NF3 workloads.
  * Added runtime validation to catch packing-geometry mismatches for NF3.

* **Tests**
  * Added a CUDA-only test covering “production shape” NF3 behavior with tier-masked experts.
  * Updated custom-op dispatch tests to match the expanded launch arguments.

### DIFF
--- a/b12x/moe/fused/w4a16/kernel.py
+++ b/b12x/moe/fused/w4a16/kernel.py
@@ -6986,6 +6986,19 @@ def run_w4a16_moe(
                 f"planned={actual_fused + (int(fused_launch.max_m_blocks),)}"
             )
         fused = fused_launch
+    if weight_layout == "nf3_2p1":
+        prepared_tiles = (
+            int(getattr(prepared, "fc1_tile_n", 0)),
+            int(getattr(prepared, "fc2_tile_n", 0)),
+        )
+        compiled_tiles = (int(fused.fc1_tile_n), int(fused.fc2_tile_n))
+        if prepared_tiles != compiled_tiles:
+            raise RuntimeError(
+                "prepared NF3 packing geometry does not match the compiled "
+                "W4A16 launch: "
+                f"prepared_fc1_fc2_tile_n={prepared_tiles}, "
+                f"compiled_fc1_fc2_tile_n={compiled_tiles}"
+            )
     capacity_m = int(fused.size_m)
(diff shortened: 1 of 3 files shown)


voipmonitor:
Current-master validation: the W4A16 launch fake-dispatch test and production-shape NF3 tier-mask test both pass (2/2). Ruff passes on all changed files. Two unrelated failures in the full launch-op file reproduce on unmodified current master and are not introduced by this PR.

voipmonitor:
Canonical-stack validation update: current `master` (`115926e`) plus #28, #32, #34, and #35 merges without conflicts. The combined SM120 suite passed: 120 passed, 1 skipped (CuTe runtime patching, W4A8 auxiliary-stream launch, W4A16/NF3 launch geometry, and NVFP4 MLA KV cache/API coverage). `git diff --check` is clean.

voipmonitor:
Final canonical-stack validation against `master@115926eb`: merged #28 -> #32 -> #34 -> #35 without conflicts, then ran the combined production-path suite: **122 passed, 1 skipped**. `git diff --check` and Ruff diagnostics on the changed files are clean. No runtime overlay is required.

PR #36, opened 2026-07-17
perf(moe): add mapped Grid188 NF3 hybrid decode kernel

voipmonitor:
## Summary

- Add the exact SM120 Grid188 one-grid MoE specialization for GLM-5.2 TP4 hybrid checkpoints with 64 NVFP4 experts and 192 NF3 experts.
- Map FC1 as 128 active CTAs plus 60 idle CTAs and FC2 as 16x5 plus 172x4 tasks, keeping exactly one resident CTA per SM.
- Validate device geometry, shared memory, registers, local memory, ABI geometry, and exact task partitions before admitting the kernel.
- Expose a graph-safe custom op and retain the existing serial kernels as the fallback for every non-exact geometry.

This is a symbol-level port of the Grid188 kernel published in davidsyoung/vllm-glm52#2; unchanged b12x code was not copied from the release overlay.

## Validation

- 7/7 focused task-map tests passed.
- Ruff check and format check passed.
- Real GLM-5.2 TP4/DCP4/MTP3 capture and inference passed on four RTX PRO 6000 Blackwell GPUs; logs confirmed the mapped kernel executed and coding output had no corruption.

## Measured result

Matched exact-checkpoint runs, three 60-second samples, CC1 decode:

| Path | Median aggregate tok/s | Mean aggregate tok/s |
|---|---:|---:|
| Existing serial fallback | 93.98 | 94.04 |
| Grid188 | 104.22 | 104.10 |

Median gain: +10.9%. The vLLM integration and fail-closed admission are submitted separately against dev/gilded-gnosis.

## Summary by CodeRabbit

* **New Features**
  * Added a hybrid W4A16 mapped-grid execution path for fused MoE workloads, including a new grid188 mapping and compile/launch interfaces.
  * Exposed APIs for exact task mapping plus mapping proof output.
* **Bug Fixes**
  * Improved validation and masking/shape-branch behavior for W4A16 fused MoE/GEMM execution.
  * Clearing caches now also clears the hybrid grid188 compile cache when present.
* **Tests**
  * Added tests to verify the grid188 mapping proof and proper errors for unsupported task counts.

### DIFF
--- a/b12x/moe/fused/w4a16/kernel.py
+++ b/b12x/moe/fused/w4a16/kernel.py
@@ -7,6 +7,7 @@ from dataclasses import dataclass, replace
 from typing import NamedTuple
 
 import cuda.bindings.driver as cuda
+import cuda.bindings.runtime as cuda_runtime
 import cutlass
 import cutlass.cute as cute
 import torch
@@ -648,9 +649,7 @@ class W4A16GemmKernel:
                     "nf3_2p1 W4A16 weights require scale_format='e4m3_k32'"
                 )
             if element_dtype != "bf16":
-                raise ValueError(
-                    "nf3_2p1 W4A16 weights are bf16-activation only (v1)"
-                )
+                raise ValueError("nf3_2p1 W4A16 weights are bf16-activation only (v1)")
         if epilogue_activation not in (None, "relu2"):
             raise ValueError(
                 "W4A16 GEMM epilogue activation currently supports only relu2"
@@ -661,9 +660,7 @@ class W4A16GemmKernel:
         has_n_tile_tail = int(size_n) % int(tile_n) != 0
         has_k_tile_tail = int(size_k) % int(tile_k) != 0
         has_scale_k_tail = int(size_k) % scale_group_size != 0
-        has_logical_tail = (
-            has_n_tile_tail or has_k_tile_tail or has_scale_k_tail
-        )
+        has_logical_tail = has_n_tile_tail or has_k_tile_tail or has_scale_k_tail
         if has_logical_tail:
             if weight_layout != "modelopt" or scale_format != "e8m0_k32":
                 raise ValueError(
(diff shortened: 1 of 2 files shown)


lukealonso:
Thanks for this port — it proved something real. The one-grid fusion win is confirmed and larger than the end-to-end number suggested: at the same GLM-5.2 TP4 shard geometry, graph-replayed on a 188-SM SM120 part, the fused hybrid runs the MoE layer in **95.9 µs vs 167.5 µs** for the serial two-launch path at m=4 (**1.75×**), and **75.3 µs vs 149.0 µs at m=1 (1.98×)**. Both of this PR's scheduling instincts also survived measurement: launching at the SM cap beats FC1-right-sizing (FC2 critical path 5 waves vs 6, ~10%), and whole-tile waves beat the split-K tail on mixed routing.

I'm closing the PR because we've landed the capability in a different shape rather than merging this implementation. The concern wasn't the kernel — it was the packaging: a sealed subclass with its own compile/launch/admission stack duplicates the persistent-loop scheduling and phase assembly, and pins parent internals it can't see change. Instead, the hybrid mode is now a trace-time specialization of the existing fused kernel:

- a tile-emission hook at `_run_persistent_gemm`'s expert-resolution seam plus a `schedule_whole_tiles` mode — every pre-existing config verified **byte-identical** before/after against a captured compile baseline;
- `W4A16FusedMoeHybridKernel` composing two per-tier fused kernels, dispatching via the same `tier<<8|local` descriptor map this PR introduced (kept bit-compatible with the integration's encoding);
- your grid/schedule choices expressed as measured launch policy in `_w4a16_fused_persistent_grid_x` rather than hardcoded 188s, so other SM120 parts adapt;
- the spill/geometry admission kept as compile-time fail-closed checks plus a CI test (122 regs/thread, zero local memory at this geometry);
- a parity suite against the serial two-launch oracle (mixed/single-tier/invalid routes, both schedules, partial batches).

One functional upside of the unified ABI: dynamic m up to the compiled capacity, so the m<4 decode buckets ride the one-grid path too — that's where the speedup peaks. The vLLM integration has been adapted accordingly on the `dev/gilded-gnosis` line with the arming/fallback flow from your companion change intact.

The b12x commit credits this PR and its origin (davidsyoung/vllm-glm52#2). Appreciate the contribution — the measured result and the schedule design both carried through.

🤖 Drafted with [Claude Code](https://claude.com/claude-code)

davidsyoung:
All gravy, thanks for porting it @lukealonso!

PR #39, opened 2026-07-18
fix(w4a16): admit ultra-wide FC2 tile in forced re-pin validation

sjug:
## Summary

W4A16 TC-decode can auto-select the ultra-wide FC2 tile
`(tile_k=32, tile_n=512)` on 48-SM GB10 geometry. Before this fix,
re-pinning that selected configuration across the custom-op boundary rejected
it because the generic tile validator enforces `tile_k >= 64`.

The result was a deterministic graph-capture boot failure for
DeepSeek-v4-Flash TP2 on DGX Spark even though sparkinfer itself had selected
the tile.

## Root cause

The ultra-wide FC2 override intentionally validates `(32, 512)` outside
`_candidate_tile_fits`, because it is the one supported TC-decode tile below
the generic `tile_k >= 64` floor. The `force_tile_config` path did not share
that exception and therefore rejected the selector's output.

## Fix

- Forward-port the fix onto the current namespaced W4A16 implementation.
- Extract one `_tc_decode_ultra_fc2_tile_fits` predicate and use it in both
  automatic selection and forced re-pin validation.
- Keep exact N/K coverage, scale-group alignment, and shared-memory capacity
  checks.
- Preserve generic validation for every other tile configuration.

## Regression coverage

`test_w4a16_tc_decode_ultra_fc2_tile_round_trips_forced_repin`:

- simulates the 48-SM GB10 wave-balance geometry;
- asserts auto-selection produces
  `(fc1_k=64, fc1_n=256, fc2_k=32, fc2_n=512)`;
- feeds that complete configuration back through `force_tile_config` and
  asserts it is preserved;
- verifies the exception remains rejected when the FC2 problem N is not
  512-aligned.

## Validation

- Before fix: exact reproduction selected `(64, 256, 32, 512)` and raised
  `ValueError` during forced re-pin.
- After fix: exact reproduction accepts the re-pinned configuration.
- `pytest -q tests/moe/test_w4a16_e2e.py::test_w4a16_tc_decode_ultra_fc2_tile_round_trips_forced_repin`
  — 1 passed.
- `pytest -q tests/moe/test_w4a16_e2e.py -k tc_decode`
  — 14 passed, 79 deselected.
- `ruff check sparkinfer/moe/_shared/kernels/w4a16/kernel.py tests/moe/test_w4a16_e2e.py`
  — passed.
- `git diff --check` — passed.

The original deployment evidence remains: on 2× GB10, DeepSeek-v4-Flash TP2
failed graph capture on every boot before the fix and booted and served
successfully with the ultra tile admitted.

### DIFF
--- a/sparkinfer/moe/_shared/kernels/w4a16/kernel.py
+++ b/sparkinfer/moe/_shared/kernels/w4a16/kernel.py
@@ -215,6 +215,8 @@ _LARGE_BATCH_TILE_CONFIGS = (
     (64, 128, 128),
     (128, 64, 128),
 )
+_TC_DECODE_ULTRA_FC2_TILE_K = 32
+_TC_DECODE_ULTRA_FC2_TILE_N = 512
 
 
 def _covering_count(total: int, quantum: int) -> int:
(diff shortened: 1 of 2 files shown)


lukealonso:

PR #43, opened 2026-07-19
fix(pcie): isolate recycled CUDA graph streams

voipmonitor:
## Problem

vLLM captures the target model, draft prefill, and draft decode in separate CUDA graph manager contexts. CUDA/PyTorch may recycle the same numeric outer and nested capture stream handles between those contexts. The PCIe oneshot and DCP A2A pools cached channels only by stream key, so a later draft capture could retain the target graph channel and race on its signal/staging buffers during replay. The existing nested-capture tests used distinct stream IDs and did not cover reuse.

A TP6/DCP6/MTP3 E2E run reproduced an Xid 13 out-of-range address at CC16; the same workload completed with CUDA_LAUNCH_BLOCKING=1, which is consistent with this stream-channel race.

## Fix

- Allocate one graph-owned channel for every independent pool capture context, even when its outer stream key was recycled.
- Make the active enclosing capture channel take precedence over stale nested stream-key aliases.
- Keep replaced graph-owned channels alive until pool shutdown, because captured graphs retain their IPC pointers.
- Apply the same ownership rule to PCIe oneshot all-reduce and PCIe DCP A2A.

## Tests

- Add target/draft regressions that deliberately reuse the same outer and nested stream IDs.
- Verify all replaced channels remain owned and are disposed on pool close.
- `34 passed, 1 skipped` across both PCIe suites plus the GPU torture test (GPU-only test skipped in the unit-test container).
- Ruff and `git diff --check` pass.

E2E validation against TP6/DCP6/MTP3 and TP8/DCP2 long-prefill configurations follows in the integration image.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved CUDA graph capture handling for PCIe communication pools.
  - Separate capture sessions now receive distinct communication channels, even when stream keys are reused.
  - Nested or active captures now consistently reuse the correct enclosing channel.
  - Pool shutdown now reliably closes all channels created during normal use and graph capture, preventing resource leaks.

- **Tests**
  - Added coverage for reused capture stream keys, channel isolation, and complete cleanup.

### DIFF
--- a/b12x/distributed/pcie_dcp_a2a.py
+++ b/b12x/distributed/pcie_dcp_a2a.py
@@ -512,6 +512,7 @@ class PCIeDCPA2APool:
         self.single_channel = bool(single_channel)
         self._channel_factory = channel_factory
         self._channels: dict[int, PCIeDCPA2A] = {}
+        self._all_channels: list[PCIeDCPA2A] = []
         self._capture_channel_stack: list[PCIeDCPA2A] = []
         self._closed = False
         if channel_factory is None and exchange_group is None:
@@ -583,6 +584,7 @@ class PCIeDCPA2APool:
                 stream_affine=not self.single_channel,
             )
         channel._bind_stream_key(stream_key)
+        self._all_channels.append(channel)
         return channel
 
     def for_stream(self, stream: object = None) -> PCIeDCPA2A:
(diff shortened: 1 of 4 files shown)


voipmonitor:
E2E validation rejected this approach. TP6/DCP6/MTP3 still reproduced Xid 13 at CC16, and forcing one fresh channel per graph-manager capture reduced CC1 from ~95 tok/s to ~30 tok/s. The unit regression modeled stream-handle reuse correctly but not the required collective ordering/lifecycle. Closing rather than leaving an unsafe change mergeable; a narrower fix will follow only after GPU validation.

PR #47, opened 2026-07-19
fix(concurrency): isolate capture channels and barrier-free NVFP4 decode

voipmonitor:
## Summary

- key recycled CUDA graph stream channels by capture ownership instead of a reused stream handle
- apply the same ownership rule to PCIe oneshot all-reduce and DCP A2A pools
- checkpoint disposable graph channels and roll them back after profiling captures
- coordinate rollback across ranks: synchronize, unmap every imported IPC handle, then free exported allocations only after a process-group barrier
- run native NVFP4 A4 micro decode as separate FC1/FC2 launches for M<=8, removing its resident-grid software barrier
- expose a plan capability telling vLLM when B12X MoE can safely overlap unrelated auxiliary-stream CUDA work
- retain `B12X_NVFP4_SPLIT_DECODE=0` as a diagnostic fallback

## Root causes

Two independent concurrency hazards surfaced under target/draft CUDA graphs and GLM shared-expert overlap. Recycled capture streams could inherit a channel selected for a prior capture owner. Separately, the fused B12X micro MoE launch uses a resident-grid software barrier; co-scheduling unrelated work can prevent every barrier participant from becoming resident.

Disposable profiling graphs add a lifetime constraint: all ranks must destroy graph users and unmap imported CUDA IPC allocations before any rank frees the corresponding exported slab. Rank-local close/free could otherwise leave a peer with a stale mapping. The rollback path now uses a three-barrier teardown protocol and restores only the pre-profile channel map.

The existing W4A8-on-NVFP4 two-phase launch already removed the micro barrier. This PR applies the same split to native NVFP4 A4 and exports the selected launch-plan capability. Dynamic and W4A16 resident launches remain correctly marked unsafe for external overlap.

## Validation

- 40 targeted B12X PCIe channel/lifecycle tests passed
- teardown tests assert the cross-rank `barrier -> close imports -> barrier -> free exports -> barrier` order for both pools
- native A4 fused versus split on the GLM TP8 expert shape: bit-identical, max/mean absolute difference 0
- native A4 CUDA graph replay with changed input and routing: bit-identical on two replays
- exact TP8/DCP2/MTP3 GLM reproducer passed short and 300k-token requests with the vLLM integration and native allocator
- Ruff and `git diff --check` passed

## Integration

Pair with local-inference-lab/vllm#131 for disposable profiling-graph cleanup and local-inference-lab/vllm#132 for stream/workspace isolation and MoE scheduling.

## Summary by CodeRabbit

- **New Features**
  - Exposed a new integration API to indicate when micro MoE plans can overlap auxiliary CUDA work.
  - Added opt-in native NVFP4 “split decode” support (eligibility rules + environment flag).
  - Added channel pool checkpoint/rollback for both PCIe DCP A2A and PCIe oneshot flows.
- **Bug Fixes**
  - Improved idempotent, ordered IPC import/export teardown and safer channel close behavior.
  - Tightened CUDA-graph capture handling and channel isolation during reused capture keys.
- **Tests**
  - Added/expanded distributed and CUDA-graph replay coverage, including rollback safety during capture and teardown resilience.

### DIFF
--- a/b12x/distributed/pcie_dcp_a2a.py
+++ b/b12x/distributed/pcie_dcp_a2a.py
@@ -19,6 +19,7 @@ from .pcie_oneshot import (
     IPC_SLAB_ALIGNMENT,
     PCIeOneshotAllReduce,
     _align_up,
+    _coordinated_close_channels,
     _current_stream_key,
     _is_current_stream_capturing,
     _normalize_device,
@@ -205,6 +206,8 @@ class PCIeDCPA2A:
         self._stream_affine = bool(stream_affine)
         self._owner_stream_key: Optional[int] = None
         self._closed = False
+        self._ipc_imports_closed = False
+        self._ipc_exports_freed = False
         self._ptr = self._ext.init_dcp_a2a(
             list(signal_ptrs),
             list(staging0_ptrs),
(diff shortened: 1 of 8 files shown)


voipmonitor:
Addressed the coordinated-teardown review finding in dee3b5e: every dispose/close/free is exception-isolated, terminal state is always recorded, cleanup remains idempotent, and a regression test verifies all resources are attempted even when every CUDA IPC operation raises. Validation: 32/32 pcie_oneshot tests passed.

voipmonitor:
Superseded by #58, rebuilt directly on current master with the namespaced sparkinfer APIs, coordinated graph-channel teardown, regression tests, and the corrected W4A16 pre-resident overlap contract.

PR #48, opened 2026-07-20
fix(moe): size W4A8 scratch for padded TP shards

voipmonitor:
## Summary

- use the repacked N128-padded extent consistently in W4A8-MX dynamic-kernel planning and launch
- size materialized activation/scale scratch and task geometry from that physical extent
- retain the logical expert width in the public execution plan
- add diagnostics and a regression test for the exact GLM-5.2 virtual-TP6 shape

## Root cause

For virtual TP6, GLM exposes a logical per-rank expert width of `N=352`. Repacked W4A8-MX weights launch a zero-padded `N=384` kernel extent. The previous planner used 352 while the launch used 384, so the dynamic kernel could exceed its preplanned materialized-intermediate scratch during startup profiling:

```text
dynamic W4A8 materialized scratch exceeds preplanned capacity
```

The fix centralizes the physical extent in `_dynamic_kernel_intermediate_size()` and uses it for tile selection, task geometry, direct-slice eligibility, scratch sizing, and launch. Other quant modes remain unchanged.

## Validation

- `ruff check sparkinfer/moe/fused_moe/_impl.py tests/moe/test_tp_moe_scratch_bindings.py`
- `pytest -q tests/moe/test_tp_moe_scratch_bindings.py`: 29 passed
- exact E2E GLM-5.2 MXFP4/A8 TP6+DCP3 startup and inference passed
- matched TP6/DCP3 MTP0 decode: 57.27 tok/s
- matched TP6/DCP3 KV capacity: 700,449 tokens at seq=16, batch=4096, GMU=0.95

This refresh rebases the original fix onto the renamed SparkInfer tree and supersedes the earlier implementation commit without changing the PR scope.

### DIFF
--- a/sparkinfer/moe/fused_moe/_impl.py
+++ b/sparkinfer/moe/fused_moe/_impl.py
@@ -1187,6 +1187,20 @@ def _is_w4a8_quant_mode(quant_mode: str) -> bool:
     return quant_mode in _W4A8_QUANT_MODES
 
 
+def _dynamic_kernel_intermediate_size(n: int, quant_mode: str) -> int:
+    """Return the intermediate extent consumed by the dynamic kernel.
+
+    Prepared W4A8-MX weights use ceil-tiled N128 storage. The launch treats
+    that zero-padded tail as part of the kernel extent, so every plan-time
+    geometry and scratch calculation must use the same padded value.
+    """
+
+    n = int(n)
+    if _normalize_quant_mode(quant_mode) == "w4a8_mx":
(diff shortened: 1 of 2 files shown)


voipmonitor:
Addressed the CodeRabbit test nitpick in 73de085: the TP6 W4A8 regression test now bounds the N128 alignment slack above as well as below. The focused test passes; the full file reports 28 additional passes and one unrelated CUDA-device failure when run in a CPU-only test container.

PR #51, opened 2026-07-20
fix(mxfp8): support tail-padded linear output

voipmonitor:
## Summary

Add an optional `tail_padding_bytes` argument to the native MXFP8 linear path and write the dense GEMM result directly into output storage with a mapped tail.

## Why

The vLLM GLM MLA path feeds some linear outputs into SM120 cuBLAS strided BMMs. The guarded repro showed cuBLAS may read up to the next 64 KiB boundary past the logical BF16 tensor end for the relevant `_v_up_proj` shape. Normal PyTorch allocations usually hide this with allocator slack; tight custom/provided buffers can fault.

This change lets vLLM request a small mapped tail for MXFP8 linear outputs without adding a separate hot-path `.clone()` / `.contiguous()` copy.

## Changes

- Add internal tail-padded output allocation for `mxfp8_linear`.
- Pass that output directly to `dense_gemm(..., out=...)`.
- Preserve existing default behavior when `tail_padding_bytes=0`.
- Apply bias in-place after GEMM so the padded storage owner is preserved.
- Extend the CUDA test to assert the tail-padding contract, including the bias path.

## Validation

- `python -m pytest -q tests/test_gemm_mxfp8_linear.py -k "test_mxfp8_linear_writes_directly_to_tail_padded_output"` in a CUDA 13.2 diagnostic container: 1 passed.
- Direct CUDA smoke test with bias: output shape `[6, 64]`, tail bytes `65536`, bitwise match against the non-tail-padded path.

## Summary by CodeRabbit

* **New Features**
  * Added optional tail padding for MXFP8 linear outputs via `tail_padding_bytes`, reserving extra output storage.
  * Works across normal execution, zero-token handling, and CUDA graph capture/replay.
  * Added input validation to reject negative `tail_padding_bytes`.
  * Bias is now applied in place on the output.
* **Tests**
  * Added assertions for tail-padded storage behavior and numerical equivalence with non-tail-padded outputs.
  * Updated CUDA graph capture/replay coverage to include tail padding.

### DIFF
--- a/b12x/gemm/mxfp8_linear.py
+++ b/b12x/gemm/mxfp8_linear.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import math
 from dataclasses import dataclass
 
 import cutlass.cute as cute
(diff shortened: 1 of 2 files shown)


lukealonso:
I'm willing to add a feature where the caller can provide their own output tensor if you want to work around this in vllm, but adding a feature to this library to compensate for a cublas bug doesn't seem right.

PR #55, opened 2026-07-21
Fix silent candidate-buffer overflow in indexer top-k

tobymao:
## Summary

`SparseNSATiledTopkKernel` selects the top-k by bucketing candidates on a radix key and storing the threshold bucket in a fixed 4096-slot shared buffer (`_SMEM_CANDS`). When a single bucket holds more than 4096 candidates — which happens on **low-contrast rows** (many near-equal scores) at longer context — the overflow is silently dropped and the lowest-indexed survivors are kept, so the paged prefill top-k (`index_topk_fp8` / `packed_contiguous`) selects tokens scoring far below the true top-k threshold. It fails silently (no error, just wrong tokens), so it surfaces only as a quality regression on long prompts, not a crash.

The failure boundary is bucket-population > 4096, not context length per se: a monotonic-score row of 16384 tokens keeps the hot bucket at 4091 and passes; 16448 and 32768 overflow it and select ~99% wrong tokens (verified by score against an fp32 reference).

## Fix

The fused decode kernel already handled this exact case with an inline *exact overflow fallback* — a candidate-buffer-free 4-round 8-bit MSD radix that re-scans the values each round filtered by the locked key prefix, so it cannot overflow. This PR extracts that fallback into a shared `@cute.jit` helper (`_exact_overflow_fallback` in `tiled_topk.py`, with a `flat` constexpr flag selecting the value loader) and routes both kernels through it:

- `SparseNSATiledTopkKernel` (and `run_row_topk`, the same kernel class backing the two-level fold) now invoke it when the threshold bucket overflows — **fixing the bug**.
- `fused_indexer.py` drops its inline copy and calls the shared helper — **removing the duplication**.

The fast (non-overflow) path is unchanged: one uniform smem load + compare per CTA gates the fallback, which runs only on the rare clustered case. Kernel cache-version strings are bumped so stale JIT binaries are invalidated.

## Testing

New regression test `tests/test_paged_prefill_topk_long_context.py` fails before this change and passes after. It checks selection by **score** against an fp32 reference (robust to ties, unlike an index match), plus **uniqueness and in-range** of the selected slots, across:

- the bucket-overflow boundary (16384 passes; 16448 and 32768 overflow),
- all-equal-score tie-fill (single bucket holds every token),
- logical-output two-level fold (`output_physical_slots=False`),
- multi-chunk carry fold (`supertile_k < context`).

Existing `test_paged_indexer_integration` (11/11) and `test_fused_indexer` continue to pass; the fused overflow-merge path is exercised directly and matches the fp32 golden.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved top-k selection reliability when candidate buffers become full.
  - Prevented valid results from being omitted during high-volume or threshold-heavy selections.
  - Added an exact fallback path to ensure deterministic, accurate candidate ranking in overflow scenarios.
  - Updated compiled kernel variants to ensure the corrected selection behavior is used consistently.

### DIFF
--- a/sparkinfer/attention/nsa_indexer/fused_indexer.py
+++ b/sparkinfer/attention/nsa_indexer/fused_indexer.py
@@ -63,6 +63,7 @@ from sparkinfer.attention.nsa_indexer.tiled_topk import (
     _THREADS_PER_CTA as _RADIX_THREADS,
     _convert_to_uint8,
     _convert_to_uint32,
+    _exact_overflow_fallback,
     _smem_ld,
     _smem_red_add,
     _smem_st,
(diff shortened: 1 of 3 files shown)


lukealonso:
LGTM - thanks.

PR #56, opened 2026-07-21
Fix deadlock in the long-context indexer tests on DGX Spark

tobymao:
## Summary

`test_fused_indexer_paged_long_context_merge_exact[96, 188]` hangs the GPU forever on a DGX Spark (GB10).

The fused indexer splits each query's work across many parallel units that then rendezvous to combine their results. Each unit occupies one GPU core, and the Spark has only 48 of them. The tests ask for 96 and 188 units, more than 48, so they can never all rendezvous at once, and the kernel waits forever.

Those numbers are fine on bigger Blackwell cards (188+ cores), which is where the test was written. On the 48-core Spark they deadlock.

## Fix

When a group needs more units than the GPU has cores, use the merge path that does not require everyone to meet at once (each unit drops its result and leaves, the last one combines them). Same result, no hang. Normal runs, which never over-split, are unchanged.

## Testing

- `test_fused_indexer_paged_long_context_merge_exact[96, 188]` now passes on a 48-core Spark (was a hard hang); still takes the fast path on bigger cards.
- Added a small unit check for the boundary, since the hang itself cannot be run as a test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented potential deadlocks during fused attention processing when workload groups exceed available GPU resources.
  * Ensured oversized groups use a safe serial processing path while preserving cooperative processing when resources allow.
  * Improved caching and selection of processing variants for consistent CUDA Graph behavior.

* **Tests**
  * Added regression coverage for cooperative processing eligibility across different GPU resource configurations.

### DIFF
--- a/sparkinfer/attention/nsa_indexer/fused_indexer.py
+++ b/sparkinfer/attention/nsa_indexer/fused_indexer.py
@@ -1171,6 +1171,7 @@ class SparseNSAFusedIndexerKernel:
         kv_layout: int = KV_LAYOUT_PAGED,
         paged_output: bool = False,
         ctas_per_group: int = 1,
+        num_sms: int = 0,
         merge_threshold: int = _LAST_CTA_MERGE_MAX,
         k_quant_page_stride: int = _PAGE_SIZE * _INDEX_HEAD_DIM,
         k_scales_row_stride: int = _PAGE_SIZE,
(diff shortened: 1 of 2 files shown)


tobymao:
hey @lukealonso thanks for merging this is in. i'd love to become more involved in this project. do you have a channel where you discuss issues and priorities?

PR #58, opened 2026-07-21
Fix PCIe graph lifecycle and preserve safe MoE overlap

voipmonitor:
## Summary

- isolate recycled CUDA graph streams and workspaces from live communication state
- roll back disposable graph channels and coordinate IPC teardown across ranks
- retain the barrier-free native NVFP4 split-decode path
- expose an explicit plan capability for shared-expert auxiliary-stream use
- allow W4A16 decode to use the safe pre-resident overlap contract for 1-7 tokens
- port the concurrency and execution-model regressions to the namespaced `sparkinfer` APIs

## Root cause

The previous stack mixed two separate concerns. Recycled graph resources needed strict ownership and coordinated teardown, while the first safety response disabled W4A16 auxiliary-stream use entirely. That blanket disable prevented resident-grid deadlocks but also removed useful overlap from the common DCP1 A16 decode path.

The corrected contract is narrower: shared-expert work may overlap gate/router work, but the consumer stream must join the auxiliary stream before submitting a resident routed-MoE grid. A positive plan capability therefore does not authorize concurrent resident-grid execution.

## Validation

- SparkInfer targeted suite: **76 passed, 6 skipped**
- paired vLLM targeted suite: **134 passed**
- DCP1 A16 decode A/B:
  - blanket W4A16 disable: **83.565 tok/s mean**
  - corrected pre-resident overlap: **86.458 tok/s mean**
  - recovery: **+3.46%**

This supersedes #47 with a clean branch directly based on `master`. The separate cuBLAS head-major safety work remains in #54.

## Summary by CodeRabbit

* **New Features**
  * Added support for checking auxiliary-stream overlap capabilities in fused MoE execution plans.
  * Improved CUDA Graph capture channel handling, including safer reuse and rollback of temporary channels.
  * Enhanced PCIe communication resource cleanup and coordinated teardown.

* **Bug Fixes**
  * Prevented stale channel reuse across captures.
  * Improved cleanup reliability when resource-release operations encounter errors.
  * Strengthened NVFP4 split-decode and auxiliary-stream behavior across supported execution modes.

* **Tests**
  * Added coverage for graph capture, rollback, coordinated cleanup, NVFP4 execution, and auxiliary-stream concurrency.

PR #59, opened 2026-07-21
fix(pcie): isolate CUDA graph channel lifecycle

voipmonitor:
## Summary

- bind recycled target and draft CUDA graph streams to independent PCIe channels
- checkpoint and roll back channels created by disposable graph captures
- coordinate IPC teardown across ranks and make rollback fail-safe
- keep all MoE execution-plan changes out of this PR

## Why

Target and speculative graph managers can reuse CUDA stream identities while retaining different live graph resources. Sharing or prematurely recycling a PCIe channel across those captures can leave stale IPC state or destroy a channel still owned by a replayable graph.

This PR contains only the PCIe lifecycle half of #58 and is based directly on `master`.

## Validation

- `tests/comm/test_pcie_dcp_a2a.py`
- `tests/comm/test_pcie_oneshot.py`
- **41 passed**
- Ruff lint and format checks passed
- split integration tree is byte-identical to the PCIe files in #58

### DIFF
--- a/sparkinfer/comm/pcie/pcie_dcp_a2a.py
+++ b/sparkinfer/comm/pcie/pcie_dcp_a2a.py
@@ -19,6 +19,7 @@ from .pcie_oneshot import (
     IPC_SLAB_ALIGNMENT,
     PCIeOneshotAllReduce,
     _align_up,
+    _coordinated_close_channels,
     _current_stream_key,
     _is_current_stream_capturing,
     _normalize_device,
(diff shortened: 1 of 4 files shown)


voipmonitor:
Manual fallback review completed because the automated review was rate-limited. The channel ownership, capture isolation, rollback ordering, and coordinated IPC teardown paths were inspected against the current master; no actionable defect was found. Focused result: 41 passed; Ruff clean.

@coderabbitai review

PR #60, opened 2026-07-21
fix(moe): expose safe shared-expert overlap plans

voipmonitor:
## Summary

- expose an explicit execution-plan capability for shared-expert auxiliary-stream scheduling
- retain bounded native NVFP4 micro-decode overlap
- reject auxiliary overlap for resident split-decode plans
- expose the safe W4A16 pre-resident capability for 1-7 token decode
- keep all PCIe graph/channel lifecycle changes out of this PR

## Why

Resident-grid MoE kernels can deadlock when auxiliary work occupies an SM required by a grid-wide barrier. The caller must query the concrete launch plan instead of treating every SparkInfer kernel as equally overlap-safe.

This PR contains only the MoE execution-model half of #58 and is based directly on `master`.

## Validation

- `tests/moe/test_moe_execution_model.py`
- `tests/moe/test_w4a8_tp_moe.py`
- CUDA image run: **35 passed, 6 skipped**
- CPU-only run: **28 passed, 13 skipped**
- Ruff lint and format checks passed
- split integration tree is byte-identical to the MoE files in #58

## Summary by CodeRabbit

* **New Features**
  * Added support for checking whether auxiliary-stream overlap is safe for MoE execution plans.
  * Added optional NVFP4 split-decode behavior, configurable through an environment setting.
* **Bug Fixes**
  * Improved stream-overlap safety for NVFP4 and W4A16 execution scenarios.
  * Verified consistent outputs when toggling NVFP4 split decode and during CUDA graph replay with auxiliary stream work.
* **Tests**
  * Expanded coverage across quantization modes, token sizes, and CUDA stream configurations.


voipmonitor:
Closing after revalidating the original root-cause evidence.

The claimed shared-expert overlap failure was not reproduced. On both 2829061 and current 695c011, M=1/4/7 pass serial eager, serial CUDA graph, routed-first overlap, and aux-first overlap. M=8 differs even in pure serial eager execution with no graph and no auxiliary stream, while heavy auxiliary work stays within the same variation range. The exact M=4 stress workload also passes on both commits, and no deadlock was observed.

Therefore the M<=7 capability boundary and caller-facing overlap policy were inferred from a pre-existing M=8 numerical/race behavior rather than an overlap-specific failure. Commit 695c011 now expresses the all-CTA admission requirement at the kernel launch through a cooperative launch, which is the appropriate layer for that contract.

The M=8 behavior is being investigated separately against real checkpoint weights and an independent oracle.

voipmonitor:
Final M=8 result from a real GLM-5.2 NVFP4 layer (K=6144, local N=256, E=256, top-k=8): this case does not execute fused micro. The production planner selects `micro` for M=7 (56 routed rows) and `dynamic` for M=8 (64 routed rows, at the strict cutover).

M=7 was bitwise identical across all repeats and matched the independent oracle at cosine 0.9999105. M=8 used the default BF16 atomic scatter of the dynamic kernel; its expected reduction-order variation was small (pairwise cosine about 0.999977, max absolute difference 0.000366) and was present in serial execution. Setting `SPARKINFER_DYNAMIC_DETERMINISTIC_OUTPUT=1` made all eight M=8 repeats bitwise identical while retaining the same oracle accuracy (cosine 0.9995970, max absolute error 0.0006714).

Therefore there is no remaining evidence of a shared-expert auxiliary-stream race or corruption in this report. The M=8 signal came from comparing different kernel families and requiring exact equality from the dynamic atomic-output path. PR #71 adds a CPU planner regression test that pins the real GLM M=7/M=8 dispatch boundary.

PR #62, opened 2026-07-21
gemm: weight-only mxfp8 batched GEMM for the MLA absorbed projections

MadeBy561:
MLA decode consumes the absorbed `W_UK_T`/`W_UV` pair, which integrations
materialize in BF16 because there is no quantized bmm — at GLM-5.2 TP4
geometry that resident pair plus its load-time dequant churn is ~724 MiB/GPU
duplicating information the kv_b mxfp8 pack already holds.

**VRAM: +724 MiB/GPU measured; max_model_len 700k → 775,000 on 4×96 GB.**

`sparkinfer/gemm/qbmm_absorb/` adds one kernel template with two orientations
serving both absorbed-projection GEMMs straight from the pack:

- `ukt`: `out[n,b,l] = sum_p a[n,b,p] * deq(pack[hs*n + p, l])` — the weight
  tile is read in native pack orientation and fed to the MMA transposed (the
  scale axis rides the GEMM N axis, constant per 32-wide window).
- `uv`: `out[n,b,v] = sum_l a[n,b,l] * deq(pack[hs*n + p_dim + v, l])` —
  layout-native e8m0-k32 row groups.

Dequant is the exact-value chain (`cvt.rn.f16x2.e4m3x2` → f32 → bf16, plus
`e<<7` e8m0 expansion with the e=0 subnormal case), bitwise identical to
`w.to(bf16) * s.view(float8_e8m0fnu).to(bf16).repeat_interleave(32, 1)`.
The e8m0 **view** is load-bearing: scales store as uint8, and a raw
`uint8.to(bf16)` converts exponent bytes as integers (127 → 127.0, not 2^0)
— a silent ×1e2–1e12 dequant catastrophe we hit in an integration-side torch
fallback during bring-up. The existing fp4-path dequant helpers are biased
variants and cannot satisfy the bitwise contract, so the module carries its
own exact-value primitives. bf16 HMMA m16n8k16, fp32 accumulators, K unsplit:
deterministic, no atomics, no workspace, no allocation on the launch path.

Serving requirements (per AGENTS.md) are load-bearing, learned on 2026-07-21:

- `warmup_qbmm_absorb()` precompiles both ops for every CUDA-graph-visible
  batch size **before any capture**. The speculator's prefill capture uses
  `num_tokens - num_reqs + 1` (= 25 at 32-token graphs / 8 seqs / k=3), which
  a target-sizes-only warmup misses: the kernel then JIT-compiles inside the
  speculator's capture and the captured draft graph replays garbage —
  observed as MTP acceptance pinned at 0.000 per position while the fully
  warmed target stayed healthy.
- A compile miss during stream capture now **raises** instead of compiling:
  a graph captured around a JIT is corrupt, and artifacts written mid-capture
  contaminate the persistent compile cache (full-tree quarantine required).

Correctness gates before performance claims (`tests/test_gemm_qbmm_absorb.py`,
sm120): one-hot bitwise dequant readout vs the bf16-chain reference; error
envelope vs fp64 no worse than the cuBLAS bf16 bmm of the materialized pair
(measured equal to 4 significant digits on every op/batch cell); replay
determinism; capture-replay equality vs eager; the capture-miss guard;
warmup covering the full batch envelope. In vivo (GLM-5.2 NVFP4+NF3 hybrid,
4×RTX 6000 Blackwell Max-Q sm120, TP4+DCP4, nvfp4 MLA KV + fp8 rope +
calibrated outer scales, MTP3): 943-row live-pack one-hot parity bitwise
clean including the MTP draft layer's pack; 400 back-to-back unsynchronized
launches (200 quiet + 200 under dual-stream contention) bitwise stable.
64-bit pool-offset rule audited: the kernel indexes weights and ≤32-row
activations, no paged pools.

Serving receipts (same rig): decode from the pack replaces the resident BF16
pair → **+724 MiB/GPU** measured post-capture; **max_model_len 700k → 775k**
on 96 GB parts (pool print 775,000; a 157,932-token single prefill runs 49 s,
flat vs the 700k rung); html-codegen decode 105.6–107.3 t/s vs 104.8–104.9
for the resident-pair baseline; MTP3 acceptance windows up to 92.1 %
(0.993/0.923/0.848 per-position), mean acceptance length 3.76.

Supersedes #61 (authored against the pre-restructure b12x layout; identical functional content, ported to the sparkinfer conventions).

## Summary by CodeRabbit

* **New Features**
  * Added support for weight-only MXFP8 batched GEMM for MLA absorbed projections.
  * Added `ukt` and `uv` operation variants with BF16 inputs and FP32-accumulated outputs.
  * Added platform and geometry support checks, warmup utilities, and CUDA Graph capture safeguards.
  * Added lazy discovery and public exports for the new GEMM operations.

* **Tests**
  * Added coverage for numerical accuracy, dequantization, determinism, warmup coverage, and CUDA Graph replay behavior.

### DIFF
--- a/sparkinfer/__init__.py
+++ b/sparkinfer/__init__.py
@@ -48,6 +48,7 @@ _OPS: tuple[str, ...] = (
     "gemm.blockscaled",
     "gemm.block_fp8_linear",
     "gemm.mxfp8_linear",
+    "gemm.qbmm_absorb",
     "gemm.wo_projection",
     "moe.fused_moe",
     "moe.ep_moe",
--- a/sparkinfer/gemm/__init__.py
+++ b/sparkinfer/gemm/__init__.py
@@ -3,6 +3,7 @@
 - ``blockscaled``: one-shot dense block-scaled GEMM (NVFP4 / MXFP4 / MXFP8).
 - ``block_fp8_linear``: DeepSeek-style serialized block-FP8 linear via MXFP8.
 - ``mxfp8_linear``: ModelOpt MXFP8 linear (one-shot).
+- ``qbmm_absorb``: weight-only MXFP8 batched GEMM for MLA absorbed projections.
 - ``wo_projection``: fused MLA WO-A/WO-B projections (+ inverse-RoPE variant).
 """
 
@@ -11,7 +12,13 @@ from __future__ import annotations
 import importlib
 from typing import Any
 
-_OP_MODULES = ("blockscaled", "block_fp8_linear", "mxfp8_linear", "wo_projection")
+_OP_MODULES = (
+    "blockscaled",
+    "block_fp8_linear",
+    "mxfp8_linear",
+    "qbmm_absorb",
+    "wo_projection",
+)
 
 
 def __getattr__(name: str) -> Any:
--- a/sparkinfer/gemm/qbmm_absorb/__init__.py
+++ b/sparkinfer/gemm/qbmm_absorb/__init__.py
@@ -1,0 +1,63 @@
+"""Weight-only MXFP8 batched GEMM for MLA absorbed projections (one-shot).
+
+Decode-path MLA consumes the absorbed projections directly from the kv_b
+mxfp8 pack (fp8 e4m3 values + per-32-in-axis e8m0 scales), replacing the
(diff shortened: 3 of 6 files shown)


lukealonso:
merged

PR #66, opened 2026-07-21
fix(dcp): support disposable A2A profiling channels

FujitsuPolycom:
## Summary

Add checkpoint/rollback lifecycle support for disposable `PCIeDCPA2A` profiling channels.

Temporary CUDA-graph profiling channels can now be discarded without leaking IPC allocations, closing channels that existed before profiling, or retaining temporary stream aliases.

## Scope

- checkpoint the pool's existing channel boundary
- roll back only channels created after that checkpoint
- validate that a checkpoint belongs to the same pool
- reject checkpoint/rollback during active CUDA graph capture
- preserve pre-existing channel and stream mappings

This is independent of the selected-record transport PRs.

## Validation

Focused tests cover successful rollback, preservation of existing channels, foreign checkpoint rejection, capture guards, stream aliases, and idempotent lifecycle behavior.

## Design reference

This lifecycle fix supports the broader profiling and graph-validation work documented in [GLM-5.2 Sparse CKV under Decode Context Parallelism](https://github.com/FujitsuPolycom/vllm/blob/codex/glm52-sparse-ckv-upstream-ready-20260721/docs/design/glm52_sparse_ckv.md).

Implementation, tests, and documentation were developed with OpenAI Codex assistance and manually reviewed.

### DIFF
--- a/sparkinfer/comm/pcie/pcie_dcp_a2a.py
+++ b/sparkinfer/comm/pcie/pcie_dcp_a2a.py
@@ -497,6 +497,12 @@ class PCIeDCPA2A:
             self.close()
 
 
+@dataclass(frozen=True)
+class _ChannelCheckpoint:
+    pool_id: int
+    channels: tuple[tuple[int, PCIeDCPA2A], ...]
+
+
 class PCIeDCPA2APool:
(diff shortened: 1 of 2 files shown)


lukealonso:

PR #71, opened 2026-07-22
test(moe): cover GLM NVFP4 dispatch boundary

voipmonitor:
## Summary

- parameterize the CPU weight-plan helper by expert count
- pin the real GLM-5.2 TP8 NVFP4 dispatch boundary in a planner test
- assert M=7 uses fused micro while M=8 uses dynamic with its default atomic-output contract

## Why

The investigation behind closed PR #60 initially treated M=8 variation as evidence about fused-micro auxiliary-stream overlap. Real checkpoint validation showed that M=8 does not execute fused micro at all: top-k 8 produces 64 routed rows, so the production planner selects the dynamic backend. Its default BF16 atomic scatter is intentionally not bitwise deterministic. M=7 remains on micro and is bitwise repeatable.

This test makes that boundary explicit so future overlap investigations do not compare two different kernel families as if they were one.

## Validation

- `pytest -q tests/moe/test_execution_model.py` in the CUDA test image: 16 passed
- real GLM layer 3, TP8/NVFP4: M=7 micro was bitwise stable; M=8 dynamic showed only expected atomic-order variation
- `SPARKINFER_DYNAMIC_DETERMINISTIC_OUTPUT=1` made all eight M=8 repeats bitwise identical

## Summary by CodeRabbit

* **Tests**
  * Added coverage for large-expert NVFP4 mixture-of-experts workloads by supporting a configurable expert count in the weight-planning helper.
  * Verified execution planning transitions from micro to dynamic mode at the expected workload size (m=8).
  * Confirmed m=8 output is non-deterministic by default, and becomes deterministic when `SPARKINFER_DYNAMIC_DETERMINISTIC_OUTPUT=1` is enabled.

### DIFF
--- a/tests/moe/test_execution_model.py
+++ b/tests/moe/test_execution_model.py
@@ -240,8 +240,12 @@ def test_workspace_plan_uses_weight_plan_source_contract() -> None:
     assert plan.spec.source_weight_scale is ScaleEncoding.E8M0_K32
 
 
-def test_glm52_tp8_nvfp4_crosses_from_micro_to_dynamic_at_m8() -> None:
+def test_glm52_tp8_nvfp4_crosses_from_micro_to_dynamic_at_m8(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
     """The GLM decode M=8 case uses dynamic's atomic output contract."""
+
+    monkeypatch.delenv("SPARKINFER_DYNAMIC_DETERMINISTIC_OUTPUT", raising=False)
 
     weights = _weight_plan(
         "nvfp4",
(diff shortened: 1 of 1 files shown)


voipmonitor:
The review finding is addressed in `4ee4cb1`. The test now removes the ambient `SPARKINFER_DYNAMIC_DETERMINISTIC_OUTPUT` value before checking the default M=8 policy, then enables it explicitly and verifies deterministic dynamic dispatch. It uses a CUDA device descriptor without allocating a GPU tensor, so the serving-only policy is exercised in CPU CI.

Validation: `tests/moe/test_execution_model.py` passes 16/16. The combined #71 + #72 integration suite passes 78/78.

PR #84, opened 2026-07-27
fix(indexer): preserve oldest threshold-boundary candidates

yatesdr:
## Summary

Adds an opt-in `oldest_boundary` selection policy to the NSA tiled top-k
indexer. The policy provides deterministic, bounds-safe handling of candidates
that share the coarse Kth-score bucket.

`exact` remains the default policy.

Closes the SparkInfer portion of
[`local-inference-lab/vllm#182`](https://github.com/local-inference-lab/vllm/issues/182).

## Background

GLM-5.2 uses a learned sparse indexer to choose 2,048 historical token
positions for attention. A token that is not selected cannot contribute to
that sparse-attention layer, so selector membership is directly observable as
long-context retrieval quality.

The reproduction uses a deterministic long document of financial-record
prose. It inserts one sentence stating that the Facility 27 maintenance ticket
number is `738216` at 40% of the document, then asks for that number at the end.
Each prompt and its chat-rendered token IDs are SHA-256 pinned. Requests use
temperature 0, thinking disabled, and a 2,000-token output limit. A request
passes only when:

- `cached_tokens == 0`;
- `finish_reason == "stop"`;
- finalized `content.strip() == "738216"`.

On the GLM-5.2 NF3/NVFP4-MLA-KV/FP8-RoPE configuration, the current v20
`exact` policy passes the pinned 245,497-token control and fails all three
pinned 343,727-token prompts. The failed requests stop normally and produce
coherent finalized output, but the target is absent.

Layer-level captures of the same 350k inputs show that the default selector
does not contain the complete three-token target value together until layer
74. With `oldest_boundary`, the complete target was selected by layer 38 and
the final answer was restored in all three cases.

## Why the policies differ

The indexer first reduces scores to a coarse radix bucket and then refines the
candidates around the Kth bucket. At long context, many positions can share
that boundary bucket. The choice of which boundary candidates enter full-score
refinement changes the 2,048-token sparse-attention set.

`exact` resolves the full quantized score field. `oldest_boundary` limits a
large tied coarse bucket by logical position before applying full-score
refinement. This produces a different, deterministic candidate set for the
learned indexer. This PR makes that choice explicit and checkpoint-selectable
without changing the default selector.

## Selection contract

With `SPARKINFER_NSA_TOPK_SELECTION_POLICY=oldest_boundary`, the selector:

1. computes an 8-bit coarse score histogram;
2. retains every candidate in a bucket strictly above the Kth bucket;
3. enumerates candidates in the Kth bucket in ascending logical-position
   order;
4. retains the first 4,096 candidates from that boundary bucket;
5. applies the existing four-pass full 32-bit radix refinement;
6. emits exactly `topk` indices.

The boundary scan uses tile-ordered warp-prefix compaction. It records the full
boundary population while writing only ordinals in `[0, 4096)`, so candidate
writes cannot exceed the allocated buffer.

## Configuration and scope

```text
SPARKINFER_NSA_TOPK_SELECTION_POLICY=exact            # default
SPARKINFER_NSA_TOPK_SELECTION_POLICY=oldest_boundary  # opt-in
```

The value is normalized and validated at module initialization. Unsupported
values fail immediately with `ValueError`.

The policy is included in the compilation-cache key. It is server-static and
must not be changed inside a running process.

This PR does not change `exact`, does not select `oldest_boundary`
automatically for other checkpoints, and does not add an overflow or
out-of-bounds compatibility path.

### Frozen end-to-end retrieval gate

All requests were cold (`cached_tokens=0`) and required finalized content
`738216` with `finish_reason=stop`.

| Cell | Default `exact` | `oldest_boundary` |
|---|---|---|
| 250k control | pass | pass |
| 350k seed 1 | target absent | pass |
| 350k seed 2 | target absent | pass |
| 350k seed 3 | target absent | pass |

The current-v20 result summary is pinned by SHA-256:

```text
dda7bddd33919d0947bcf45e0731c7fe07e1d4918944781fca9928cafe1d18f6
```

### Randomized cold ladder

The same current-v20 process passed 50k, 150k, 250k, 300k, 350k, and 475k.
The deepest prompt contained 466,493 tokens. Every cell:

- reported `cached_tokens=0`;
- returned finalized content `738216`;
- stopped normally;
- passed arithmetic, coherence, and degeneration checks.

The ladder summary is pinned by SHA-256:

```text
b855f1febae880a6ae146797fbf37707e3ea02bccd213578d41ec5ba19ae6268
```

### Selector and capacity checks

- Captured selector rows reconstruct at 97.0%--99.3% set overlap.
- A model-free 3,072-row production-shape test
  (`lengths=1..3072`, `topk=2048`, `block_q=32`, `block_k=256`) reports the
  correct valid count for every row and no out-of-range index.
- The current-v20 integration image serves a 480,000-token maximum with a
  545,280-token KV pool and zero restarts.
- Unit tests cover the default, normalization, both supported values, and
  rejection of unsupported values.

### Matched KLD comparison

Three runs per policy were compared with the same pinned BF16 reference logits
(2,047 token positions):

| Policy | Runs | Mean KLD ± SD | Min | Max |
|---|---:|---:|---:|---:|
| `exact` | 3 | 0.15823696 ± 0.00468419 | 0.15539664 | 0.16364348 |
| `oldest_boundary` | 3 | 0.16044075 ± 0.00297924 | 0.15700885 | 0.16236257 |

The paired `oldest_boundary - exact` deltas were `-0.00663463`,
`+0.00655420`, and `+0.00669180`. Their mean was `+0.00220379` (+1.39%
of the exact-policy mean), with sample SD `0.00765461`. The mixed signs and
variance larger than the mean delta do not show a large shallow
distribution-level regression at n=3.

This is a 2,048-token no-regression gate, not a selector-sensitive comparison:
the selector budget is also 2,048, so both policies select every eligible
token. Deep-context efficacy is covered by the frozen 350k gate and randomized
50k–475k ladder.

## Qualification status

The long-context retrieval, selector-safety, and KV-capacity gates pass on the
current v20 base (`vLLM 0c79e41db4`, SparkInfer `e603f74bb6`). The matched
three-run shallow KLD comparison passes without a large regression.
Prefill/decode performance comparison remains integration-qualification work
and is not claimed as complete by this PR.

The pinned image, A/B configuration, test procedure, and evidence links are
available in the
[community validation specification](https://gist.github.com/yatesdr/a2e84aa3171ee0b355649704f04f96a8).

### DIFF
--- a/sparkinfer/attention/nsa_indexer/tiled_topk.py
+++ b/sparkinfer/attention/nsa_indexer/tiled_topk.py
@@ -36,15 +36,47 @@ from sparkinfer._lib.intrinsics import (
     st_shared_i32,
 )
 from sparkinfer._lib.utils import current_cuda_stream
+from sparkinfer.attention._shared.cute import ops as cute_ops
 
 _THREADS_PER_CTA = 1024
 _DEFAULT_TOPK = 2048
 _SUPPORTED_TOPK = (512, 1024, 2048)
 _RADIX = 256
+_SELECTION_POLICY_ENV = "SPARKINFER_NSA_TOPK_SELECTION_POLICY"
+_SELECTION_POLICY_EXACT = "exact"
+_SELECTION_POLICY_OLDEST_BOUNDARY = "oldest_boundary"
+_SELECTION_POLICIES = frozenset(
+    {
+        _SELECTION_POLICY_EXACT,
+        _SELECTION_POLICY_OLDEST_BOUNDARY,
+    }
+)
+
+
+def _resolve_selection_policy(value: str | None) -> str:
+    policy = (value or _SELECTION_POLICY_EXACT).strip().lower()
+    if policy not in _SELECTION_POLICIES:
+        raise ValueError(
+            f"{_SELECTION_POLICY_ENV} must be one of "
+            f"{sorted(_SELECTION_POLICIES)}, got {policy!r}"
+        )
+    return policy
+
+
+_SELECTION_POLICY = _resolve_selection_policy(
+    os.environ.get(_SELECTION_POLICY_ENV)
+)
+_OLDEST_BOUNDARY = _SELECTION_POLICY == _SELECTION_POLICY_OLDEST_BOUNDARY
 # Use every SM120 CTA lane for the first histogram.  The four-times narrower
 # bucket keeps ordinary 32k/64k low-contrast rows on the buffered exact path;
 # truly degenerate buckets still use the rescan fallback below.
-_COARSE_RADIX_BITS = 10
+#
+# ``oldest_boundary`` gives the required positional behavior an explicit
+# contract: all candidates in strictly higher coarse buckets remain eligible,
+# while the threshold bucket is refined over its oldest 4096 virtual positions
+# in stable index order. No candidate write may cross the allocation. The
+# default remains exact.
+_COARSE_RADIX_BITS = 8 if _OLDEST_BOUNDARY else 10
 _COARSE_RADIX_BINS = 1 << _COARSE_RADIX_BITS
 _HIST_SLOTS = _COARSE_RADIX_BINS + 128
 # A 1024-thread selector CTA is already limited to one resident block per SM on
@@ -52,7 +84,7 @@ _HIST_SLOTS = _COARSE_RADIX_BINS + 128
 # long-context overflow without reducing occupancy; the complete shared-memory
 # allocation remains below the 99 KiB opt-in block limit.  The exact rescan below
 # still covers wider or degenerate threshold buckets.
-_SMEM_CANDS = 8192
+_SMEM_CANDS = 4096 if _OLDEST_BOUNDARY else 8192
 _SCAN_UNROLL = 4
 _SUPERTILE_K_ENV = "SPARKINFER_NSA_TOPK_SUPERTILE_K"
 _SUPERTILE_K_DEFAULT = 32768
@@ -539,6 +571,7 @@ class SparseNSATiledTopkKernel:
         # or the user's final output.
         self.is_first = bool(is_first)
         self.output_physical_slots = bool(output_physical_slots)
+        self.oldest_boundary = bool(_OLDEST_BOUNDARY)
 
     @cute.jit
     def __call__(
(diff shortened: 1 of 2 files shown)


yatesdr:
The digest-pinned review image and community A/B package are now public.

Image:

```text
ghcr.io/yatesdr/glm52-serve:gilded-gnosis-v20-oldest-boundary-pr84-20260727
ghcr.io/yatesdr/glm52-serve@sha256:43e5a48781ee5cf40a92cc494749b21306b72280bd1a875721a45422323f2599
```

Public package:

https://github.com/users/yatesdr/packages/container/package/glm52-serve

Community validation spec, same-image `exact`/`oldest_boundary` Compose,
frozen long-context reproducers, and fail-closed KLD wrapper:

https://gist.github.com/yatesdr/a2e84aa3171ee0b355649704f04f96a8

The first valid BF16-reference KLD smoke completed:

| Policy | Mean `KL(BF16 reference || candidate)` | Positions |
|---|---:|---:|
| `exact` | 0.16541040 | 2,047 |
| `oldest_boundary` | 0.16384826 | 2,047 |

Both runs used the same image, checkpoint, 368-byte NVFP4+FP8-RoPE KV
format, TP4/DCP1 eager posture, and the published 2,048-token BF16 logits.
Both matched the reference token IDs and emitted the fail-closed completion
marker without OOM/fatal errors.

The single-pair delta is `-0.00156214`, within historical run variation.
This is intentionally classified as a shallow no-regression result: the
reference context and selector budget are both 2,048, so this cell is not
selector-sensitive and does not prove the 350k fix. The frozen 350k gate and
randomized 475k ladder carry that causal evidence. The three-repeat KLD result
is in progress and will supersede these provisional smoke values.

Published artifact SHA-256:

```text
spec        ef682801a1c59d2f9b15c75ed2e0b4753f6cc48e8c56d92a8d4b8cc784a91398
compose     5dce77c527791069b793c835a62fe0699f6a60af37ebabbc96dd71644dd22ceb
KLD wrapper 28050af936b625b7ebd97fbe3dd778f33dfeafc0492074da1e328e4fc06a5f44
summarizer  1692595340392fc13fee2be22091241fff7004051e87d0c25b42df14380844ad
```

lukealonso:
**Naming feedback on the new selection policy (\`SPARKINFER_NSA_TOPK_SELECTION_POLICY\`)**

Neither \`exact\` nor \`oldest_boundary\` clearly conveys the actual behavioral difference:

- \`exact\` is misleading as a name for the *default* policy — its everyday tie-break behavior at the threshold bucket is non-deterministic (atomic-race order among GPU threads); it's only "exact" in the sense that it falls back to a precise 4-pass rescan when the bucket overflows the candidate buffer. Something like \`race\` / \`atomic\` (with \`exact_fallback\` behavior called out separately) would better reflect what it does day-to-day.
- \`oldest_boundary\` bundles two separate properties into one name: (1) deterministic/stable tie-breaking, and (2) the specific tie-break rule (prefer smallest/earliest index). A name like \`stable_oldest\` or \`deterministic_oldest\` would disentangle "this is reproducible" from "this is the direction of the bias," and would also avoid reading like a cache-eviction policy name.

**Correctness concern: crowded threshold buckets under \`oldest_boundary\`**

For the new policy, the coarse radix bin is narrower (8-bit vs 10-bit) and the candidate buffer is halved (4096 vs 8192), *and* the exact-overflow rescan fallback that exists for the default policy is disabled. This means that if the threshold bucket ends up with more than 4096 tied/qualifying candidates (more likely given the coarser binning), the kernel silently keeps only the oldest 4096 by stable index order and drops the remainder — producing top-k results that are provably incorrect relative to the true top-k, with no fallback or error signal. Given how much more crowding a coarser bin can produce, this failure mode seems more reachable than the equivalent overflow case under the default policy. Worth documenting explicitly (or emitting a warning/counter) so callers relying on \`oldest_boundary\` are aware they're trading correctness for determinism in dense-tie scenarios.

yatesdr:
The digest-pinned review image and community A/B package are now public.

Image:

```text
ghcr.io/yatesdr/glm52-serve:gilded-gnosis-v20-oldest-boundary-pr84-20260727
ghcr.io/yatesdr/glm52-serve@sha256:43e5a48781ee5cf40a92cc494749b21306b72280bd1a875721a45422323f2599
```

Public package:

https://github.com/users/yatesdr/packages/container/package/glm52-serve

Community validation spec, same-image `exact`/`oldest_boundary` Compose,
frozen long-context reproducers, and fail-closed KLD wrapper:

https://gist.github.com/yatesdr/a2e84aa3171ee0b355649704f04f96a8

The matched three-run BF16-reference KLD comparison completed:

| Policy | Runs | Mean `KL(BF16 reference || candidate)` ± SD | Min | Max |
|---|---:|---:|---:|---:|
| `exact` | 3 | 0.15823696 ± 0.00468419 | 0.15539664 | 0.16364348 |
| `oldest_boundary` | 3 | 0.16044075 ± 0.00297924 | 0.15700885 | 0.16236257 |

All six runs used the same image, checkpoint, 368-byte NVFP4+FP8-RoPE KV
format, TP4/DCP1 eager posture, and the published 2,048-token BF16 logits.
Every run matched the reference token IDs and emitted the fail-closed
completion marker without OOM/fatal errors.

Paired `oldest_boundary - exact` deltas were `-0.00663463`, `+0.00655420`,
and `+0.00669180`. The mean delta was `+0.00220379` (+1.39% of the
exact-policy mean), with sample SD `0.00765461`. The mixed signs and variance
larger than the mean do not show a large shallow distribution-level regression.
This is intentionally classified as a shallow no-regression result: the
reference context and selector budget are both 2,048, so this cell is not
selector-sensitive and does not prove the 350k fix. The frozen 350k gate and
randomized 475k ladder carry that causal evidence.

Published artifact SHA-256:

```text
spec           af70f2043bea50d3d50aadbf48ad875601a04a30eb97304fc9e8e8b1b2d839df
compose        5dce77c527791069b793c835a62fe0699f6a60af37ebabbc96dd71644dd22ceb
matrix wrapper 1ba0c392953baf21c11b1a36190680c767f249fe5f4ad826df47b5c638d178ad
summarizer     fc81b049872bf19aebd9d0e449c64d52f58cee446210fea58c5716370b217dbc
aggregate JSON e9bc81d775e6830b8b3101943a6bba714b50ffd65aab8255c4a5d07acb663ff3
summary        1d18c949f4ccf7f053fe4188dd2aecbf05fefef0dcdab9a2ca2c9d3321647ce1d
full report    a5e0faabf73064ec6435bad02b03bd54fe1cf45c73e7c2c44ae9627c1e6821bc
```

yatesdr:
Superseded by SparkInfer #86 and vLLM #189, which offer an improved root-cause fix.

PR #92, opened 2026-07-30
moe: stabilize Trellis arena memory profiling

voipmonitor:
## Problem

The unified Trellis `fused_moe` path used the generic W4A16 compile bucket for its caller-owned arena and left several fixed launch variants to first use. For a 3,072-token EXL3 prefill plan this rounded the arena to 4,096 rows, reserving about 320 MiB/rank that execution could not use. First-use compilation during vLLM profiling also produced non-repeatable rank-local activation peaks and therefore non-repeatable distributed KV capacity.

## Changes

- Preserve exact fixed capacity for `exl3_trellis_mcg` scratch plans; generic W4A16 plans retain compile bucketing.
- Compile fixed Trellis fused-MoE and top-k-sum variants during plan creation.
- Prewarm mapped and unmapped route packing before memory profiling.
- Retain strong references to launch objects but leave runtime binding unresolved, preserving the faster live-M resolver used by the unified API.
- Add exact-capacity and prewarm/dispatch contract tests.

The original Trellis implementation from closed #90 is already present on `master`; this PR contains only the follow-up profiling and capacity fix.

## Validation

- SparkInfer Trellis regression set: 124 passed
- Paired vLLM worker tests: 3 passed
- TP4/DCP4/MTP0, three cold starts: 834,560 KV tokens every time, 1.62 GiB activation peak on every rank
- MTP0 decode: 48.00 and 48.58 tok/s, no errors
- TP4/DCP4/MTP3: 485,888 KV tokens, 101.02 tok/s, acceptance 0.6847
- `git diff --check`: clean

Paired vLLM PRs: local-inference-lab/vllm#190 and local-inference-lab/vllm#198.

## Summary by CodeRabbit

* **Performance Improvements**
  * Enhanced Trellis FP4 (W4A16, full-rotation) planning by precompiling and caching fused and top‑k-sum launches for reuse.
  * Improved W4A16 workspace/token capacity handling for more consistent scratch-plan sizing.
  * Reduced unnecessary runtime dispatch during scratch-plan binding by using preplanned launches.

* **Bug Fixes**
  * Tightened fixed-capacity workspace sizing to match expected Trellis geometry and prevent under-provisioning.

* **Tests**
  * Added CUDA Trellis coverage for launch planning and scratch-plan prewarming, including validation across expected launch and mapping variants.

PR #100, opened 2026-07-30
moe: bound W4A16 scratch and repair tail correctness

malaiwah:
## Summary

Bound unpinned W4A16 scratch planning to route block sizes reachable within the declared token capacity, mirror the correction in replicated-input EP, and expose a cheap public arena-sizing query.

This is stacked on #92 because the fixed-capacity Trellis planning API under review there is the source base. It addresses the safe planner work from #93; it does **not** clamp Trellis capacity or implement caller-side prefill chunking.

Related: #93

## Problem

`fused_moe` and `ep_moe` sized unpinned W4A16 route buffers and both fp32 `c_tmp` planes across every allowed block size `(8, 16, 32, 48, 64)`, even when runtime selection cannot reach most of them within `max_tokens`.

At `max_tokens=64`, `topk=8`, `E=160`, SM count 188, hidden 6144, intermediate/rank 512, the current full sweep prices the non-full W4A16 arena at 94.94 MiB. Runtime can select only block 8, whose exact arena is 30.03 MiB: **64.91 MiB/rank of unreachable reserve**.

The review also exposed two sizing hazards that a naive narrowing would make visible:

1. packed direct-topk/TC-decode uses `m * topk * block_m` scratch geometry instead of packed-route geometry;
2. non-Trellis binding cannot carry an expert map, so core route packing always uses `weight_E`, even when the router workspace uses a smaller `route_num_experts`.

Both are covered explicitly rather than relying on the old oversized sweep.

## Change

- Add `route_block_sizes_for_capacity(max_tokens, topk, num_experts)` using the same monotone selector as runtime.
- Size unpinned fused and EP route buffers only across that conservative reachable interval.
- Size fused `c_tmp` independently for both packed-route and every reachable direct-topk/TC-decode `(m, block)` geometry.
- Use `weight_E` for every non-Trellis W4A16 core route allocation; retain requested `route_E` for mapped full-rotation Trellis.
- Preserve explicit block pins exactly.
- Export `fused_moe.required_nbytes(Caps)`, which returns the complete caller-owned arena requirement without launch compilation or retained storage.

## Safety invariants

- Kernel selection, launch geometry, route packing, and numerical math are unchanged.
- Every runtime-selected block for `1 <= m <= max_tokens` is included.
- Block-8 doubled fp32 scratch accounting remains intact.
- EP still reserves power-of-two route-pack capacity before applying the reachable block set.
- Direct-topk affects only `c_tmp`; packed route-index and block-expert buffers are not inflated by direct geometry.
[...]

PR #102, opened 2026-07-30
fix(moe): reject unresolved capture launches

### LINKED ISSUE #98: [bug] W4A16 fused MoE: runtime launch resolve has no CUDA-graph capture guard, so a first-use CuTe compile can run inside a captured graph
Found during a source review of the v20 r12 release image `voipmonitor/vllm:gilded-gnosis-v20-vllmb46c3aa-si35aebc6-fi801d57a-cu132-20260730-r12`. All cited sparkinfer files are byte-identical to public commit `de04125508c75e748016493f851e6bd4513b515b` (`git diff de04125 r12-composed -- <path>` empty for `kernel.py`, `fused_moe/_impl.py`, `_lib/compiler.py`, `w4a16/route_pack.py`, `w4a16/host.py`; composed tree `35aebc6`), so permalinks use that SHA. vLLM caller refs are from the composed r12 vLLM tree (commit `b31667d3`, tree `b46c3aa`) and are quoted inline without permalinks.

Deployment context: GLM-5.2 753B MoE (hidden 6144, 75 MoE layers, top-k 8), EXL3 TR3 3.0 bpw, TP4, `cudagraph_mode=FULL_AND_PIECEWISE`; launcher defaults `VLLM_EXL3_TRELLIS_MAX_M=32`, `VLLM_EXL3_TRELLIS_BLOCK_M=8`, `VLLM_EXL3_PREFILL_TRELLIS=1`, `VLLM_EXL3_PREFILL_BLOCK_M=64`, `MAX_BATCHED_TOKENS=3072`.

**This is a defensive-hardening report, filed with an honest reachability verdict: the hazard is LATENT on both shipped configs, not live.** It is filed anyway because (1) the compile entry point demonstrably sits *inside* the captured region on the shipped EXL3 path, (2) every layer of protection is external to sparkinfer while every sibling first-use site in the same file self-defends, (3) the one in-process backstop is bypassed on a disk-cache hit, and (4) the fix as originally proposed is wrong at the proposed location.

## Where

`compile_w4a16_fused_moe` is consulted on **every** `run_w4a16_moe` call, including from inside a captured CUDA graph, and has no capture guard: [`sparkinfer/moe/_shared/kernels/w4a16/kernel.py#L10246-L10259`](https://github.com/local-inference-lab/sparkinfer/blob/de04125508c75e748016493f851e6bd4513b515b/sparkinfer/moe/_shared/kernels/w4a16/kernel.py#L10246-L10259)

```python
    if int(prepared.workspace.numel()) < sms * 4 + 2:
        raise ValueError("prepared W4A16 workspace is too small for fused FC1+FC2")
    if fused_launch is None:
        fused = compile_w4a16_fused_moe(
            size_m=m,
            hidden_size=hidden_size,
            ...
            moe_block_size=block_size_m,
            max_m_blocks=int(required_m_blocks),
```

`fused_launch` is *always* `None` on the vLLM path — [`sparkinfer/moe/fused_moe/_impl.py#L937-L959`](https://github.com/local-inference-lab/sparkinfer/blob/de04125508c75e748016493f851e6bd4513b515b/sparkinfer/moe/fused_moe/_impl.py#L937-L959) (`TPMoEScratchPlan.bind`):

```python
            route_expert_map=route_expert_map,
            output_expert_map=output_expert_map,
            fused_launch=None,
            topk_sum_launch=None,
        )
```

The whole of `compile_w4a16_fused_moe` (kernel.py:7026-7731) contains no capture check. Exhaustive listing of `is_current_stream_capturing` in `kernel.py` — **three** occurrences, none in either compile entry point nor at the resolve site:

```
8446         if torch.cuda.is_current_stream_capturing():   # rotation placeholder
9179     if torch.cuda.is_current_stream_capturing():       # _get_c_tmp scratch
9439         if torch.cuda.is_current_stream_capturing():   # Trellis dense storage
```

Contrast the planner, which does guard — [`sparkinfer/moe/fused_moe/_impl.py#L6183-L6186`](https://github.com/local-inference-lab/sparkinfer/blob/de04125508c75e748016493f851e6bd4513b515b/sparkinfer/moe/fused_moe/_impl.py#L6183-L6186):

```python
    if not torch.cuda.is_available():
        raise RuntimeError("CUDA must be available before planning Trellis MoE")
    if torch.cuda.is_current_stream_capturing():
        raise RuntimeError("Trellis launch planning cannot run during capture")
```

and the sibling first-use compile sites elsewhere in the tree that all self-defend: `gemm/_shared/mxfp8_bmm.py:1393` and `:1490` (raise on "compile miss during CUDA-graph capture"), `gemm/mla_query_projection/_bf16.py:199-203`, `gemm/bf16_gemv/_kernel.py:289-292` (falls back to cuBLAS under capture), `w4a16/route_pack.py:96`, `moe/ep_moe/_impl.py:106`, `moe/fused_moe/_impl.py:5900` and `:5964`. The largest and slowest compile in the MoE stack (the fused FC1→act→FC2 megakernel plus two GEMM kernels) is the one hot-path first-use site with no guard.

## Mechanism

1. `bind()` hard-codes `fused_launch=None`, so the launch is resolved *lazily inside the kernel call* rather than being carried on the binding. The captured graph body therefore executes `compile_w4a16_fused_moe(...)` as Python on every MoE layer.
2. The requested cache key is `("w4a16_fused_moe", device, kernel.__cache_key__)` (kernel.py:7423-7428). `size_m` is *not* in the key (it is `dataclasses.replace`d onto a hit at L7430-7435), but `max_m_blocks`, `moe_block_size`, `cta_m_blocks`, `blocks_per_sm`, `tile_n/tile_k` and `shared_words` of both GEMMs are (kernel.py:1000-1037, nested via L4810-4811).
3. `max_m_blocks` comes from `required_m_blocks = int(block_expert_ids.numel())` (kernel.py:10201), which `pack_topk_routes_by_expert` slices to `capacity_route_blocks` derived from `route_pack_numel_capacity(numel, topk)` → `route_pack_token_capacity` → `1 << (tokens-1).bit_length()` (`host.py:232-240`, `route_pack.py:307-331`, `route_pack.py:108-112`). So the key is a deterministic function of the **next-pow2 token bucket** of `m`, plus host-side constants.
4. On a `_FUSED_CACHE` miss (`kernel.py:6606`, an unbounded module-level `dict`) the function proceeds straight into `sparkinfer_compile(...)` (kernel.py:7659, 8146, 8236) → `sparkinfer._lib.compiler.compile`, i.e. an NVVM/ptxas JIT plus CUDA module load, with a CUDA graph capture in progress.

## Reachability (LATENT on both shipped configs)

I tried to construct a live path and could not. Recorded in full so the margin is auditable:

**The entry point is genuinely inside capture.** Shipped EXL3 MoE routes `exl3.py:_apply_rank_sliced` → `api.bind(plan, scratch=...)` → `TPMoEScratchPlan.bind` (`fused_launch=None`) → `run_w4a16_moe`, for every decode step, inside the FULL decode graphs. Not eager-only.

**But vLLM warms every key first.** `GPUModelRunner._warmup_and_capture` (`vllm/v1/worker/gpu_model_runner.py:6913-6938`) runs `cudagraph_num_of_warmups` eager `_dummy_run`s at *exactly* `desc.num_tokens` immediately before capturing that descriptor, and `vllm/config/vllm.py:1430` forces `cudagraph_num_of_warmups = 1` whenever cudagraphs are enabled (the `CompilationConfig` default at `config/compilation.py:643` is `0`). The FULL path adds a second eager re-warm (`vllm/v1/worker/gpu/cudagraph_utils.py:500-509`), and `CUDAGraphWrapper` optionally a third (`vllm/compilation/cuda_graph.py:317-329`). Since the key is a pure function of `m` = `desc.num_tokens` plus constants, and `_FUSED_CACHE` is unbounded (no eviction between warmup and capture), **every key capture requests was already populated eagerly.**

**Recomputed key coverage** (top-k 8, E=160 local experts, `max_m_blocks(m) = ceil(max_packed_route_slots(next_pow2(m)*topk, bs, E) / bs)`; capture sizes are `MTP+1 .. GRAPH` in steps of `MTP+1`, `serve-glm52-v16.sh:337-345`):

| capture m | pow2 bucket | plan | `bs` | `max_m_blocks` |
|---|---|---|---|---|
| 4 | 4 | decode | 8 | 32 |
| 8 | 8 | decode | 8 | 64 |
| 12, 16 | 16 | decode | 8 | 128 |
| 20, 24, 28, 32 | 32 | decode | 8 | 172 |
| 36…64 (`m > TRELLIS_MAX_M=32`) | 64 | prefill | 64 | 166 |

Five distinct GEMM keys for a `GRAPH=32` capture set; six with `GRAPH=64`. Per-size eager warmup is therefore *load-bearing*, not incidental — nothing else populates these.

**Second line of defence, also external.** Both capture sites wrap `torch.cuda.graph(...)` in `guard_b12x_kernel_resolution` (`vllm/compilation/cuda_graph.py:332-341` for PIECEWISE, `vllm/v1/worker/gpu/cudagraph_utils.py:510-516` for FULL), which calls `sparkinfer.freeze_kernel_resolution`. It is active on the shipped configs because `serve-glm52-v16.sh:227-239` unconditionally exports `VLLM_USE_B12X_WO_PROJECTION=1`, `VLLM_USE_B12X_MHC=1`, `VLLM_USE_B12X_SPARSE_INDEXER=1` (plus `VLLM_USE_B12X_MOE=1` for `MOE_BACKEND=b12x`), satisfying `b12x_cuda_graph_prewarm_enabled()`. A genuine `cute.compile` inside capture therefore hits `raise_if_kernel_resolution_frozen` (`_lib/compiler.py:2727` / `:2777`) → `KernelResolutionFrozenError`.

**Where the margin is thin — three independent ways it disappears:**

- **The freeze does not cover a disk-cache hit.** `_lib/compiler.py:2659-2679` returns the loaded object *before* `raise_if_kernel_resolution_frozen` is ever reached:
  ```python
      if disk_cache_enabled:
          compiled = _load_cute_compile_from_disk(cache_key)
          if compiled is not None:
              ...
              _memory_cache_put(memory_cache_key, compiled)
              return compiled
  ```
  `SPARKINFER_COMPILE_DISK_CACHE` defaults to `"1"` (`_lib/compiler.py:655-657`) and the image pins a persistent `SPARKINFER_COMPILE_CACHE_DIR=/cache/jit/${CACHE_FINGERPRINT}/sparkinfer/compile` (`Dockerfile.vllm-b12x-cu132:1234`). So on a warm host cache a first-touch-under-capture takes `_load_cute_compile_from_disk` → `tempfile.TemporaryDirectory` + `shutil.copy2` + `ExternalBinaryModule(...)` (a CUDA module load) *inside the capture region*, silently, with no diagnostic at all.
- Protection (i) is a **vLLM upstream default**, not a sparkinfer invariant: `cudagraph_num_of_warmups` is `0` in `CompilationConfig`, and any embedder of sparkinfer that captures without an eager pass at the same `m` has zero protection.
- Protection (ii) is gated on `VLLM_USE_B12X_*`; a config with all of them `0` silently drops the freeze.

## Failure mode if it fires

Two sub-cases, honestly separated:

- **Disk-cache miss:** `raise_if_kernel_resolution_frozen` fires first (when the freeze is active) → clean `KernelResolutionFrozenError` before any CUDA work. Without the freeze: a multi-hundred-ms-to-seconds NVVM/ptxas JIT plus module load inside the capture window.
- **Disk-cache hit:** module load inside capture, freeze bypassed. Whether `ExternalBinaryModule` construction performs work that CUDA rejects during stream capture (and hence whether `cudaStreamEndCapture` fails loudly or a subtly broken graph is produced) **is not statically provable from this tree** — `ExternalBinaryModule` lives in `cutlass.base_dsl.export`, outside both repos. Measurement needed (see Verification below). I am not claiming silent corruption.

The project already treats compile-inside-capture as a defect class, but only in offline validation, never at runtime: `validation/cutlass_migration/acceptance/w4a16_serving.py:416-418` (`"graph capture compiled a CUTLASS kernel"`), `.../compute_exceptions.py:748-749`, `.../residual_composite.py:558-560`, `.../diagnostics/paired/w4a16_serving.py:820`.

## Severity

`bug` — defensive. **No silent numeric corruption is demonstrated, and no live path exists on the shipped gate or turnkey configs.** The impact is a missing invariant on the one unguarded first-use compile that provably executes inside captured graphs, whose entire safety today is supplied by the host framework. Cost when it fires ranges from a loud startup error (best case) to an unaudited CUDA module load inside a capture region (disk-hit case). No memory cost, inside or outside the profiled budget.

## Suggested fix

**Not** the obvious one. Mirroring the planner's guard *inside* `compile_w4a16_fused_moe` would be wrong: `_prewarm_w4a16_planned_launches` deliberately calls it while capturing — [`_impl.py#L6519-L6520`](https://github.com/local-inference-lab/sparkinfer/blob/de04125508c75e748016493f851e6bd4513b515b/sparkinfer/moe/fused_moe/_impl.py#L6519-L6520) captures `is_capturing = torch.cuda.is_current_stream_capturing()` and then at `#L6625-L6643` skips **only** the route-pack launch under capture, keeping the `compile_w4a16_fused_moe` call at `#L6565` on purpose ("The real route-pack launch happens in `run_w4a16_moe`. During CUDA graph capture this prewarm-only launch would be recorded as useless work"). A guard in the compile function would break that supported path.

Correct placement is the **runtime resolve site**, where the miss is unintended:

```python
    if fused_launch is None:
        if torch.cuda.is_current_stream_capturing():
            key_probe = _fused_moe_cache_probe(...)   # existing key, no compile
            if key_probe is None:
                raise RuntimeError(
                    "W4A16 fused MoE launch is not resolved for CUDA graph capture "
                    f"(m={m}, moe_block_size={block_size_m}, "
                    f"max_m_blocks={required_m_blocks}); run an eager warmup at this "
                    "token count before capturing"
                )
        fused = compile_w4a16_fused_moe(...)
```

Raising is right here, not falling back: unlike `_get_c_tmp` (kernel.py:9164-9184, which raises under capture and allocates in eager) and unlike `bf16_gemv/_kernel.py:290` (which has a legitimate cuBLAS fallback), there is **no alternative MoE launch** to fall back to — the only options are compile or fail, and compiling is what we are trying to prevent. It is also capture-safe (pure host-side `dict` probe, no CUDA API call, no allocation) and changes neither numerics nor any ABI. Cheapest correct variant: cache-probe first and raise only on a genuine miss, so the guard never rejects a warm key.

Two adjacent items, deliberately out of scope here:
- `_lib/compiler.py:2659-2679` should consult `raise_if_kernel_resolution_frozen` before the disk-cache load, not after, so `freeze_kernel_resolution` means what its name says.
- Bucket mismatch corroborating #96 part B: the plan-time prewarm computes `capacity_m_blocks` from `capacity_tokens * num_topk` **exactly** (`_impl.py:6219-6226`) while the runtime resolve uses the pow2 bucket. Recomputed at the turnkey prefill capacity (`m=3072`, `bs=64`, `topk=8`, `E=160`): planner `capacity_m_blocks = 542`, runtime `required_m_blocks = 670` (bucket 4096). So the prewarmed key is unreachable from the runtime probe by *two* independent mechanisms, not just the tile-config pin.

## Verification / repro hint

1. **Prove the entry point is inside capture** (no GPU needed beyond a normal boot): set `SPARKINFER_LOG_CUTE_COMPILES_AFTER_ENGINE_START=1` and boot the turnkey; every `compile_w4a16_fused_moe` disk-hit/miss during `Capturing CUDA graphs` is logged with its cache key.
2. **Prove the latency of the warmup dependency**: launch with `--compilation-config '{"cudagraph_num_of_warmups":0,...}'` (overriding `config/vllm.py:1430`) plus `VLLM_USE_B12X_MOE=0 VLLM_USE_B12X_MHC=0 VLLM_USE_B12X_WO_PROJECTION=0 VLLM_USE_B12X_SPARSE_INDEXER=0` to drop the freeze, and `SPARKINFER_COMPILE_DISK_CACHE=0` to force a true compile. Capture should then attempt a CuTe compile for the largest capture size. Observe whether `cudaStreamEndCapture` fails or a graph is produced — this is the measurement that resolves the "silent broken graph vs. loud error" question above.
3. **Unit-level**: the harness pattern already exists — assert `compile_cache_info()["compile_misses"]` is unchanged across a capture, as `validation/cutlass_migration/acceptance/w4a16_serving.py:405-418` does; extend it to the `fused_launch=None` bind path.

## Provenance of the observation

First noted (unfiled) as "*Latent capture hazard*" in a review comment on the unrelated perf issue #96: https://github.com/local-inference-lab/sparkinfer/issues/96#issuecomment-5132287753 — parent issue #96 (W4A16 launch-path per-call cost / unreachable prewarm) is linked for context only; this guard gap is not covered by #96's body.

malaiwah:
## Summary

Make every W4A16 first-use resolution path fail before compilation, disk-cache loading, or CUDA module loading when the selected launch stream is being captured.

Fixes #98.

Stacked on #92 because the fixed-capacity Trellis planning API under review there is the source base.

## Change

- Freeze compiler resolution before persistent-cache payload construction or disk lookup while preserving in-memory hits.
- Add cache-only W4A16 fused-launch resolution for runtime capture misses.
- Detect capture on an explicitly selected non-current CUDA stream with `cuStreamIsCapturing`, while retaining the current-stream fast path.
- Use one canonical route-capacity/key derivation across planner prewarm, route packing, preplanned slicing, and runtime resolution. The 3072-token generic case therefore resolves the same 4096-token bucket and `max_m_blocks=670` everywhere; fixed-capacity Trellis remains exact.
- Keep explicit planning/prewarm allowed outside runtime launch resolution, including planning invoked while a graph capture context is active.

## Invariants

- Memory-cache hits during capture return the already-resolved launch.
- A miss fails before small-M dispatch, CuTe JIT, persistent disk lookup, filesystem/module load, or compilation.
- Non-capture behavior, launch ABI, kernel selection, and numerics are unchanged.
- Runtime capture checks the stream that will receive the launch, not merely whichever stream is current in PyTorch.

## Verification

CPU-isolated against the pulled r12 image dependencies:

```text
29 passed, 87 skipped
```

Suites:

- `tests/_lib/test_compile_cache.py`
- `tests/moe/test_w4a16_e2e.py`

Changed-file Ruff lint passes. Focused regressions cover frozen memory hits and disk-hit bypass, selected-stream capture, generic capacity bucket boundaries, exact Trellis capacity, and capture-time planned prewarm.

No live-GPU result is claimed. The active production deployment was not rebuilt or restarted.

### DIFF
--- a/sparkinfer/_lib/compiler.py
+++ b/sparkinfer/_lib/compiler.py
@@ -2648,6 +2648,15 @@ def compile(
     compiled = _memory_cache_get(memory_cache_key)
     if compiled is not None:
         return compiled
+    from sparkinfer._lib.runtime_control import (
+        raise_if_kernel_resolution_frozen,
+    )
+
+    raise_if_kernel_resolution_frozen(
+        "cute.compile",
+        target=func,
+        cache_key=compile_spec if compile_spec is not None else memory_cache_key,
+    )
 
     post_engine_start_log = _cute_compile_post_engine_start_log_enabled()
     payload = _compile_disk_cache_payload(
@@ -2720,15 +2729,6 @@ def compile(
 
             with _MEMORY_CACHE_LOCK:
                 _COMPILE_MISSES += 1
-            from sparkinfer._lib.runtime_control import (
-                raise_if_kernel_resolution_frozen,
-            )
-
-            raise_if_kernel_resolution_frozen(
-                "cute.compile",
-                target=func,
-                cache_key=compile_spec if compile_spec is not None else payload,
-            )
             call_kwargs = {
                 k: v for k, v in kwargs.items() if k != "__dsl_compile_options_key"
             }
@@ -2770,15 +2770,6 @@ def compile(
 
     with _MEMORY_CACHE_LOCK:
         _COMPILE_MISSES += 1
-    from sparkinfer._lib.runtime_control import (
-        raise_if_kernel_resolution_frozen,
-    )
-
-    raise_if_kernel_resolution_frozen(
-        "cute.compile",
-        target=func,
-        cache_key=compile_spec if compile_spec is not None else payload,
-    )
     call_kwargs = {k: v for k, v in kwargs.items() if k != "__dsl_compile_options_key"}
     compiled = _call_cute_compile(
         compile_callable,
--- a/sparkinfer/moe/_shared/kernels/dynamic.py
+++ b/sparkinfer/moe/_shared/kernels/dynamic.py
@@ -822,7 +822,11 @@ class MoEDynamicKernelBackend:
             source_tile_m=materialized_source_tile_m,
             deterministic_output=bool(deterministic_output),
             num_topk=self.num_topk,
-            activation=self.activation,
+            # This helper is gated-only and is never launched unless the split
+            # materialized path is active.  Use a valid inert specialization for
+            # non-split activations (notably ReLU2) instead of rejecting them
+            # during otherwise valid monolithic-kernel construction.
+            activation=self.activation if self.w4a8_split_materialized else "silu",
         )
         self.materialized_phase2_kernel = W4A8MaterializedPhase2Kernel(
             source_tile_m=materialized_source_tile_m,
--- a/sparkinfer/moe/_shared/kernels/dynamic.py
+++ b/sparkinfer/moe/_shared/kernels/dynamic.py
@@ -4200,15 +4200,7 @@ class MoEDynamicKernelBackend:
             # region; A bufs in sA past the FC2-intermediate bytes; FC1
             # residual bufs in sC (dead until the epilogue rendezvous); FC2
             # residual bufs alias the (FC1-only) A-buf region.
-            if cutlass.const_expr(self.w4a8_repacked):
-                w4a8_sa0 = (
-                    ctrl_base_addr
-                    + Int32(Storage._offsets["sA"])
-                    + Int32(self.tile_shape_mnk[0] * self.tile_shape_mnk[2])
-                )
-                w4a8_res0 = ctrl_base_addr + Int32(Storage._offsets["sC"])
-                w4a8_res2 = w4a8_sa0
-            elif cutlass.const_expr(self.w4a8_small):
+            if cutlass.const_expr(self.w4a8_repacked or self.w4a8_small):
                 w4a8_sa0 = (
                     ctrl_base_addr
                     + Int32(Storage._offsets["sA"])
--- a/sparkinfer/moe/_shared/kernels/w4a16/host.py
+++ b/sparkinfer/moe/_shared/kernels/w4a16/host.py
@@ -240,6 +240,27 @@ def route_pack_token_capacity(tokens: int, topk: int) -> int:
     return 1 << (max(int(tokens), 1) - 1).bit_length()
 
 
+def route_pack_capacity(
+    numel: int,
+    block_size: int,
+    num_experts: int,
+    *,
+    topk: int = 1,
+    bucket_tokens: bool = True,
+) -> tuple[int, int, int]:
+    """Return the canonical routed-row, packed-slot, and block capacities."""
+    numel_capacity = (
+        route_pack_numel_capacity(numel, topk=topk)
+        if bucket_tokens
+        else max(int(numel), 1)
+    )
+    packed_routes = max_packed_route_slots(
+        numel_capacity, int(block_size), int(num_experts)
+    )
+    route_blocks = (packed_routes + int(block_size) - 1) // int(block_size)
+    return max(numel_capacity, 1), max(packed_routes, 1), max(route_blocks, 1)
+
+
 def max_w4a16_route_capacity(routed_rows: int, num_experts: int) -> tuple[int, int]:
     route_slots = 0
     route_blocks = 0
@@ -424,6 +445,7 @@ __all__ = [
     "plan_w4a16_buffers",
     "reorder_w13_to_gate_up",
     "route_pack_numel_capacity",
+    "route_pack_capacity",
     "route_pack_token_capacity",
     "select_route_block_size_m",
     "unswizzle_block_scale",
--- a/sparkinfer/moe/_shared/kernels/w4a16/kernel.py
+++ b/sparkinfer/moe/_shared/kernels/w4a16/kernel.py
@@ -89,7 +89,6 @@ from sparkinfer.moe._shared.kernels.w4a16.host import (
     max_packed_route_slots,
     packed_gemm_scratch_elements,
     plan_w4a16_buffers,
-    route_pack_capacity,
     select_route_block_size_m,
     validate_activation,
 )
@@ -10181,21 +10180,15 @@ def run_w4a16_moe(
         )
 
     route_slots_for_scratch = int(m) * int(topk) * int(block_size_m)
-    if use_direct_topk_routes:
-        required_m_blocks = int(m) * int(topk)
-    else:
-        _, _, required_m_blocks = route_pack_capacity(
-            int(m) * int(topk),
-            int(block_size_m),
-            route_num_experts,
-            topk=topk,
-        )
+    required_m_blocks = int(m) * int(topk) if use_direct_topk_routes else 0
     if fused_launch is not None and not use_direct_topk_routes:
-        _, route_slots_capacity, route_blocks_capacity = route_pack_capacity(
+        route_slots_capacity = max_packed_route_slots(
             int(fused_launch.size_m) * int(topk),
             int(block_size_m),
             route_num_experts,
-            topk=topk,
+        )
+        route_blocks_capacity = (route_slots_capacity + int(block_size_m) - 1) // int(
+            block_size_m
         )
         if packed_route_indices is not None:
             if int(packed_route_indices.numel()) < route_slots_capacity:
@@ -10231,12 +10224,7 @@ def run_w4a16_moe(
             )
         )
         route_slots_for_scratch = int(packed_route_indices.numel())
-        if int(block_expert_ids.numel()) != int(required_m_blocks):
-            raise RuntimeError(
-                "W4A16 route packing did not produce the canonical launch capacity: "
-                f"runtime={int(block_expert_ids.numel())}, "
-                f"planned={int(required_m_blocks)}"
-            )
+        required_m_blocks = int(block_expert_ids.numel())
 
     props = torch.cuda.get_device_properties(a_input.device)
     sms = int(props.multi_processor_count)
--- a/sparkinfer/moe/_shared/kernels/w4a16/kernel.py
+++ b/sparkinfer/moe/_shared/kernels/w4a16/kernel.py
@@ -7054,6 +7054,7 @@ def compile_w4a16_fused_moe(
     full_rotation: bool = False,
     rotation_input_dtype: str | None = None,
     broadcast_suh: bool = False,
+    _require_cached: bool = False,
 ) -> W4A16FusedMoeCompileResult:
     scale_format = _normalize_scale_format(scale_format)
     intermediate_rotation = bool(intermediate_rotation)
@@ -7432,6 +7433,13 @@ def compile_w4a16_fused_moe(
             size_m=size_m,
             max_m_blocks=max_m_blocks,
             blocks_per_sm=kernel.blocks_per_sm,
+        )
+    if _require_cached:
+        raise RuntimeError(
+            "W4A16 fused MoE launch is not resolved for CUDA graph capture "
+            f"(m={size_m}, moe_block_size={moe_block_size}, "
+            f"max_m_blocks={max_m_blocks}); run an eager warmup at this token "
+            "count before capturing"
         )
 
     if (not collect_activation_amax) and _small_m_direct_supported(
@@ -10274,6 +10282,7 @@ def run_w4a16_moe(
             intermediate_rotation=intermediate_rotation_scales is not None,
             full_rotation=full_rotation,
             rotation_input_dtype=rotation_input_dtype,
+            _require_cached=torch.cuda.is_current_stream_capturing(),
         )
     else:
         if int(fused_launch.size_m) < m:
--- a/sparkinfer/moe/_shared/kernels/w4a16/kernel.py
+++ b/sparkinfer/moe/_shared/kernels/w4a16/kernel.py
@@ -89,6 +89,7 @@ from sparkinfer.moe._shared.kernels.w4a16.host import (
     max_packed_route_slots,
     packed_gemm_scratch_elements,
     plan_w4a16_buffers,
+    route_pack_capacity,
     select_route_block_size_m,
     validate_activation,
 )
@@ -9745,6 +9746,23 @@ def _resolve_route_block_size_m(
     return compiled
 
 
+def _w4a16_stream_is_capturing(
+    stream: cuda.CUstream,
+    *,
+    current_stream: cuda.CUstream,
+) -> bool:
+    """Observe capture on either Torch's current stream or an explicit stream."""
+    current_capturing = torch.cuda.is_current_stream_capturing()
+    if current_capturing or int(stream) == int(current_stream):
+        return current_capturing
+    result, status = cuda.cuStreamIsCapturing(stream)
+    if result != cuda.CUresult.CUDA_SUCCESS:
+        raise RuntimeError(
+            f"cuStreamIsCapturing failed for the selected W4A16 stream: {result}"
+        )
+    return int(status) != 0
+
+
 def run_w4a16_moe(
     a_input: torch.Tensor,
     prepared,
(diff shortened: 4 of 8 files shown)


malaiwah:
## Rental-machine validation — repaired composition now PASS

Tested PR head: `2bed880a7e9edbd9f2d976ba1a8ee88c9ba6e338`  
Tested base: `06032ce74afba49a683d01339ed2c971568cec3f` (PR #92 branch)

Host: Vast.ai 46335896; 4x RTX PRO 6000 Blackwell, all `NODE`/NUMA 0.  
Driver/CUDA/PyTorch: 610.43.02 / CUDA UMD 13.3 / PyTorch 2.12.0+cu132.

### Isolated result

The handoff head `5bea8f3c1c04…` failed the GPU `relu2` case because it instantiated the inactive gated helper. Validation added:

- `2c2e22010073` — avoid the inactive gated helper for `relu2`;
- `2bed880a7e9e` — combine identical W4A8 branches / Ruff cleanup.

The current head passed:

- 29 CPU tests, 87 skipped; Ruff clean;
- 108 executing GPU tests, 8 skipped, cold and warm;
- capture hit/miss guards, non-current captured stream handling, generic 3072→4096 resolution, and exact fixed Trellis 3072 coverage.

### Composition chronology

The first clean #100 + #102 merge failed two ordered GPU cases with NaN oracle metrics after scratch reuse. This was a real integration blocker, not an isolated #102 failure.

After repairing #100's exposed scratch-lifetime and tail-bound defects, exact composition `cae3aad137bd739881833c064e1912177698156f` now passes:

- **220 passed, 16 skipped** cold in 171.98 s and warm in 5.91 s;
- poisoned eager and 16-replay graph coverage;
- **24/24 Compute Sanitizer cases, 0 errors**;
- byte-identical output/allocation/peak-memory in a matched three-run A/B.

The prior composition blocker is resolved. Full chronology and raw evidence: [repair ledger at `d4b0a90`](https://github.com/malaiwah/glm52-exl3-vast/blob/d4b0a90/docs/field-review-results/2026-07-30-vast-46335896/UPSTREAM-REPAIR-CAMPAIGN.md#b-sparkinfer-w4a16-scratch-lifetime).

### Exact r14 deployable-source replay

The immutable r14 SparkInfer release tree was independently reconstructed from `blackwell-llm-docker@2464cc03`: base `9b852b281250123fe323f63ccb1df3cac0f3bbca` plus the release lock produced the published tree `8110e3ea417794bfb08aff1fba20135102e5536b` exactly. Applying the accepted #100/#102 repairs produced candidate `e2205cba8d78db03427a3945a75a1f647b51ef5a` (tree `1d378d599538daecfc1e8c11a44010b27d3ddfe5`).

On physical GPU 2 with a fresh candidate-specific extension cache, the complete seven-file union passed **332 passed / 16 intentional skips** cold in 237.03 s and identically warm in 10.03 s. This closes the image-source composition boundary for the W4 repair; no new numerical, compile-cache, capture, or mixed-K regression appeared.

Exact source provenance and immutable raw logs: [r14 reconstruction ledger](https://github.com/malaiwah/glm52-exl3-vast/blob/fd1960b/docs/field-review-results/2026-07-30-vast-46335896/UPSTREAM-REPAIR-CAMPAIGN.md#f-exact-r14-base-reconstruction), [cold log](https://github.com/malaiwah/glm52-exl3-vast/blob/fd1960b/docs/field-review-results/2026-07-30-vast-46335896/artifacts/sparkinfer-r14-w4-e2205cba-gpu2-cold.log), [warm log](https://github.com/malaiwah/glm52-exl3-vast/blob/fd1960b/docs/field-review-results/2026-07-30-vast-46335896/artifacts/sparkinfer-r14-w4-e2205cba-gpu2-warm.log).

malaiwah:
Final cross-stack evidence for exact head 2bed880a7e9edbd9f2d976ba1a8ee88c9ba6e338: 29 CPU tests passed, 108 executing GPU tests passed cold and warm, and the repaired #100 plus #102 composition passed 220 tests/16 skips plus 24/24 Compute Sanitizer cases with zero errors. Capture misses fail before small-M dispatch, persistent-cache lookup, module load, or compilation, while pre-resolved hits remain valid. The complete GLM gate establishes normal serving compatibility; the targeted negative tests, not a warmed serving run, prove the first-use capture-miss guard. Evidence: https://github.com/malaiwah/glm52-exl3-vast/blob/5c76a2536e7fc9a5f1cb6bf182531889f5385e65/docs/field-review-results/2026-07-30-vast-46335896/UPSTREAM-REPAIR-CAMPAIGN.md

malaiwah:
## Review-base note

This PR still targets `fix/trellis-stable-profile-20260730`, whose #92 content reached `master` through a non-ancestral merge/squash. I checked the graph before changing the base: a naive retarget is **not patch-neutral**. It changes the visible series from 5 to 8 commits and adds `tests/moe/test_tp_moe_scratch_bindings.py` to the hypothetical three-dot diff.

I am therefore leaving the tested head/base untouched instead of manufacturing a misleading clean check. Reviewers should use exact head `2bed880a7e9edbd9f2d976ba1a8ee88c9ba6e338` and the immutable manifest/patch digest in the counter-validation guide. A clean `master` replay should be a separately range-diffed successor, not an in-place base flip.

Guide: https://github.com/malaiwah/glm52-exl3-vast/blob/d821001f97dd15f0ae6ace626cc1f334a2069553/docs/field-review-results/2026-07-30-vast-46335896/COUNTER-VALIDATION.md

voipmonitor:
Superseded by #108. The original PR was stacked on an obsolete pre-#92 base. #108 contains only the unresolved-capture guard commits directly on current `master`, preserving the focused tests and avoiding replay of already-merged history.

PR #108, opened 2026-08-01
fix(moe): reject unresolved capture launches

voipmonitor:
## Summary

Reject unresolved W4A16 launch resolution before compilation, persistent-cache lookup, filesystem/module loading, or CUDA module loading when the selected launch stream is being captured.

This is the current-`master` successor to #102. It contains only the capture-resolution commits after the obsolete stacked base.

## Change

- Preserve already-resolved in-memory launch hits during capture.
- Freeze resolution before persistent-cache payload construction or disk lookup.
- Add cache-only W4A16 fused-launch resolution for runtime capture misses.
- Detect capture on an explicitly selected non-current CUDA stream with `cuStreamIsCapturing`, while keeping the current-stream fast path.
- Use one route-capacity/key derivation across planner prewarm, route packing, slicing, and runtime resolution.
- Keep explicit eager planning and prewarm legal outside runtime launch resolution.

The failure is early and actionable: the operator must run eager warmup for the missing token count before capture. Non-capture behavior, launch ABI, kernel selection, and numerics are unchanged.

## Verification

```text
29 passed, 87 skipped
```

Focused suites cover memory hits, disk-hit rejection, selected-stream capture, generic capacity buckets, exact Trellis capacity, and capture-time planned prewarm. Changed-file Ruff checks and `git diff --check` pass.

No merge is performed by this PR creation.

## Summary by CodeRabbit

* **New Features**
  * Improved mixture-of-experts workload planning for more efficient token routing and packing.
  * Added support for a broader range of activation functions in optimized execution paths.
  * Improved compatibility with CUDA Graph capture and selected execution streams.

* **Bug Fixes**
  * Prevented unexpected compilation during graph capture.
  * Improved behavior when kernel resolution is frozen while cached kernels remain available.
  * Enhanced cache planning consistency for variable workloads.

### DIFF
--- a/sparkinfer/_lib/compiler.py
+++ b/sparkinfer/_lib/compiler.py
@@ -2648,6 +2648,15 @@ def compile(
     compiled = _memory_cache_get(memory_cache_key)
     if compiled is not None:
         return compiled
+    from sparkinfer._lib.runtime_control import (
+        raise_if_kernel_resolution_frozen,
+    )
+
+    raise_if_kernel_resolution_frozen(
+        "cute.compile",
+        target=func,
+        cache_key=compile_spec if compile_spec is not None else memory_cache_key,
+    )
 
     post_engine_start_log = _cute_compile_post_engine_start_log_enabled()
     payload = _compile_disk_cache_payload(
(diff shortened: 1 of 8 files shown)


voipmonitor:
Superseded by #110. Although #108 and #109 were individually mergeable against `master`, either merge order conflicted in their shared E2E test file. #110 preserves both commit series, resolves the single mechanical import conflict by retaining both imports, and passes the combined 89-test portable suite plus Ruff formatting.

PR #109, opened 2026-08-01
moe: bound W4A16 scratch and repair tail correctness

voipmonitor:
## Summary

Bound unpinned W4A16 scratch to launch geometries reachable within the declared token capacity, expose a cheap arena-size query, and repair the tail cases that the tighter planner made reproducible.

This is the current-`master` successor to #100. It contains only the commits after that PR's obsolete stacked base; no already-merged Trellis history is replayed.

## Planner changes

- Compute the conservative set of route block sizes reachable for `1 <= m <= max_tokens`.
- Size fused and replicated-input EP route buffers only for that set.
- Account independently for packed-route and direct-top-k/TC-decode `c_tmp` geometry.
- Preserve explicit block pins and mapped full-rotation Trellis behavior.
- Export `fused_moe.required_nbytes(Caps)` without compiling or retaining a launch.

For the documented `max_tokens=64`, `topk=8`, `E=160`, `H=6144`, `I/rank=512` case, the unpinned non-full W4A16 arena falls from 94.94 MiB to 30.03 MiB per rank. This does not alter kernel selection or numerical math.

## Correctness repairs

The bounded arena exposed pre-existing generic W4A16 tail defects. This PR also:

- clears or masks stale direct-FC2 scratch lanes;
- preserves native ModelOpt E8M0 scale strides and non-64 scale tails;
- bounds weight, scale, and activation tail loads by logical rows;
- chooses FC1 chunk counts that partition the logical intermediate into whole 16-value blocks;
- covers narrow and wide non-aligned residue classes in fused and EP paths.

Aligned production shapes retain their existing geometry.

## Verification

```text
32 passed, 80 skipped
```

Focused suites cover fused/EP planning, route packing, replay, non-aligned native shapes, and W4A16 E2E cases. Changed-file Ruff checks and `git diff --check` pass.

No merge is performed by this PR creation.

## Summary by CodeRabbit

* **New Features**
  * Added `required_nbytes` to estimate fused MoE workspace requirements without launching computation.
  * Improved automatic routing capacity and block-size selection for varied workloads.

* **Bug Fixes**
  * Improved handling of non-aligned dimensions and padded scale layouts.
  * Prevented out-of-bounds loads, stale intermediate data, and NaN results in W4A16/NVFP4 operations.

* **Tests**
  * Expanded coverage for workspace planning, routing capacities, bounds safety, scale padding, and non-aligned model dimensions.

PR #110, opened 2026-08-01
moe: harden W4A16 planning, tails, and capture resolution

voipmonitor:
## Summary

Combine two compatible W4A16 hardening series on current `master`:

1. bound unpinned scratch planning to reachable launch geometries and repair generic native tail cases;
2. reject unresolved launch resolution before compilation or persistent-cache/module loading during CUDA graph capture.

This supersedes #108 and #109. Those PRs were independently mergeable against `master`, but either merge order conflicted in their shared E2E test file. This PR preserves both commit series, resolves the single mechanical import conflict by retaining both imports, and adds one final repo-version Ruff formatting commit. Runtime changes merged without manual conflict resolution.

## Planner and tail correctness

- Size unpinned fused and EP route/scratch buffers only for block sizes reachable within `max_tokens`.
- Account for packed-route and direct-top-k/TC-decode geometry independently.
- Preserve explicit block pins and mapped full-rotation Trellis behavior.
- Expose `fused_moe.required_nbytes(Caps)` without compiling a launch.
- Mask stale direct-FC2 lanes and logical activation/weight/scale tails.
- Preserve native ModelOpt E8M0 scale strides and valid non-64 shapes.
- Choose FC1 chunk counts that partition logical intermediates into whole 16-value blocks.

For the documented 64-token planning case, the non-full W4A16 arena falls from 94.94 MiB to 30.03 MiB per rank. Aligned production shapes retain their established launch geometry.

## Capture contract

- Existing in-memory launch hits remain legal during capture.
- Unresolved launches fail before disk-cache lookup, filesystem/module load, or compilation.
- Capture is detected on the explicitly selected stream, including a non-current stream.
- Planner prewarm, route packing, slicing, and runtime resolution share one capacity/key derivation.
- Eager planning and prewarm remain legal outside runtime launch resolution.

## Verification

Combined focused suite on the r16 dependency image:

```text
89 passed, 259 GPU-gated skips
```

The suite covers compiler cache behavior, fused/EP planning, Trellis, route packing, replay, non-aligned native shapes, and W4A16 E2E cases. Ruff 0.15.12 lint and format, plus `git diff --check`, pass across all 16 changed Python files.

No merge is performed by this PR creation.

### DIFF
--- a/sparkinfer/_lib/compiler.py
+++ b/sparkinfer/_lib/compiler.py
@@ -2648,6 +2648,15 @@ def compile(
     compiled = _memory_cache_get(memory_cache_key)
     if compiled is not None:
         return compiled
+    from sparkinfer._lib.runtime_control import (
+        raise_if_kernel_resolution_frozen,
+    )
+
+    raise_if_kernel_resolution_frozen(
+        "cute.compile",
+        target=func,
+        cache_key=compile_spec if compile_spec is not None else memory_cache_key,
+    )
 
     post_engine_start_log = _cute_compile_post_engine_start_log_enabled()
     payload = _compile_disk_cache_payload(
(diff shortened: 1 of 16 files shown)


voipmonitor:
Superseded by the self-contained, current-master integration in #112. The successor retains this PR’s planning, tail, and capture safeguards together with the paired-M8 and qualified prefill work required by r20.

PR #115, opened 2026-08-03
[HH] Fix Kimi K3 DCP and native E8M0 graph safety

voipmonitor:
## Summary

- make the PCIe DCP head gather accept FP8 E4M3 with the correct 16-byte transaction width
- skip invalid empty-shard partial outputs when their LSE weight is zero, avoiding IEEE `0 * NaN`
- select the native E8M0 K/32 scale loader before the W4A16 packed-E4M3 path for aligned Kimi K3 FC1 segments
- add DCP graph/replay coverage, native-MXFP4 oracle coverage, and model-free dense-MLA/TP16 decode probes

Paired HH vLLM integration: local-inference-lab/vllm#232.

## Root cause

Kimi K3 uses `K=3584`, split into seven aligned FC1 segments. The generic aligned path checked W4A16 first and interpreted native E8M0 K/32 scales as packed E4M3 K/16 scales. That doubled the expert scale stride and could read outside the allocation for high expert IDs. The DCP path also treated FP8 gathers as eight-byte packs and could accumulate undefined empty-shard output even when its softmax weight was zero.

## Validation

- SparkInfer selected suite: 37 passed, 3 skipped
- arbitrary-route lifecycle, CUDA graph replay, and targeted native-E8M0 oracle checks pass
- full original Kimi K3 MXFP4 checkpoint: TP16/DCP8, SparkInfer dense MLA, FP8 KV, physical 1M cache
- three idle 1,024-token decode runs: 40.949 / 41.401 / 41.167 tok/s; 41.167 tok/s median
- coherent output with no CUDA, nonfinite, assertion, or repeated-`!` failure

Full reproduction and profiling notes: https://github.com/local-inference-lab/rtx6kpro/blob/docs/kimi-k3-hh-dense-mla-dcp8-20260803/models/kimi-k3/hh-dense-mla-full-mxfp4-dcp8-2026-08-03.md

### DIFF
--- a/benchmarks/benchmark_kimi_k3_dense_mla_splits.py
+++ b/benchmarks/benchmark_kimi_k3_dense_mla_splits.py
@@ -1,0 +1,173 @@
+#!/usr/bin/env python3
+"""Benchmark capture-static Kimi-K3 dense-MLA split plans on one GPU.
+
+This isolates the exact TP16/DCP8 per-rank production shape without loading
+the model.  Global context length is eight times ``cache_seqlens`` because a
+DCP8 rank owns one eighth of the KV cache.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import statistics
+
+import torch
+
+from sparkinfer.attention import dense_mla
+
+
+FP8 = torch.float8_e4m3fn
+HEADS = 48
+HEAD_DIM = 576
+VALUE_DIM = 512
+
+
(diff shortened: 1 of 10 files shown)


voipmonitor:
Superseded by #116, which stops at production commit a844630 and excludes the follow-up benchmark/profiling-only commit f39c6bf. The full validation branch remains preserved as the deployed/reproduced artifact.

PR #117, opened 2026-08-03
fix(trellis): make mixed expert counts runtime-dynamic

voipmonitor:
## Summary

Mixed Trellis K3/K4 expert counts and shared-H rotation flags are checkpoint
data, not compile geometry. This PR passes both through the runtime ABI while
preserving one compiled kernel for compatible 256-expert partitions.

The first commit retains Michel Belleau's original #114 change with authorship
preserved. The follow-up fixes the benchmark's storage contract, carries the
broadcast-H metadata through the mixed wrapper/cache, and adds focused
regression coverage.

## Problem

The shared-H GLM-5.2 EXL3 3.42 bpw checkpoint changes its tier partition by
layer:

- layer 3: 206 K3 + 50 K4 experts;
- layers 4-77: 148 K3 + 108 K4 experts;
- layer 78: 256 K3 experts.

Legacy checkpoints also use other valid partitions, including 192/64.
Embedding these counts or the H-storage layout in a cached kernel can reuse
stale bounds, offsets, or strides when the next layer differs.

The old benchmark also materialized only 16 experts per tier while advertising
a 192/64 launch and padded W13 only. ABI 5 correctly rejected its tight W2 and
global-scale storage before binding raw pointers.

## Changes

- pass K3/K4 counts as runtime kernel arguments;
- pass per-tier broadcast-H flags through the mixed wrapper and cache;
- derive tier views and offsets from runtime metadata;
- keep the compile cache key partition-agnostic where geometry is unchanged;
- validate expert-sized W13, W2, and global-scale storage before launch;
- make the benchmark materialize its complete declared storage contract;
- add regression coverage for dynamic partitions, shared-H, cache reuse,
  storage bounds, and CUDA graph replay.

## Validation

- Mixed-Trellis SM120 suite: 15 passed, including CUDA graph capture/replay,
  changing partitions, shared-H flags, high expert IDs, and storage rejection.
- Exact layer-3/rank-0 3.42 checkpoint benchmark:
  `206/50`, shared-H on both tiers, zero relative error, cosine 1.0.
- Mixed kernel: 70.43 us versus 110.38 us for serial K3 + K4 launches
  (`1.567x`).
- Clean TP4 runtime validation on the exact checkpoint passed DCP1/MTP0,
  DCP1/MTP3, and DCP4/MTP3, including CUDA graphs and c8/c16 batching.
- Legacy 3.25 bpw checkpoint still selects its qualified 192/64 path and boots
  with CUDA graphs.

## Relationship to #114

This supersedes #114 because its external head cannot be amended here. The
runtime-count change is retained and credited through its original commit; the
additional commits supply the shared-H contract and validation required for a
mergeable implementation.

### DIFF
--- a/benchmarks/benchmark_mixed_trellis.py
+++ b/benchmarks/benchmark_mixed_trellis.py
@@ -239,6 +239,38 @@ def _pad_projection_major_w13(tier, compiled_experts: int):
     return replace(tier, w13=padded.view(-1), num_experts=int(compiled_experts))
 
 
+def _pad_exact_tier_storage(tier, compiled_experts: int):
+    """Materialize the full expert-sized storage contract used by ABI 5."""
+    materialized = int(tier.num_experts)
+    bits = int(tier.trellis_bits)
+    padded = _pad_projection_major_w13(tier, compiled_experts)
+    w2 = torch.zeros(
+        (
+            int(compiled_experts),
+            INTERMEDIATE // 16,
+            HIDDEN // 16,
+            8 * bits,
+        ),
+        dtype=torch.int32,
+        device=tier.w2.device,
+    )
+    w2[:materialized].copy_(tier.w2.view(materialized, *w2.shape[1:]))
+
+    def pad_global(source: torch.Tensor) -> torch.Tensor:
+        output = torch.ones(
+            int(compiled_experts), dtype=source.dtype, device=source.device
+        )
+        output[:materialized].copy_(source)
+        return output
+
+    return replace(
+        padded,
+        w2=w2.view(-1),
+        w13_global_scale=pad_global(tier.w13_global_scale),
+        w2_global_scale=pad_global(tier.w2_global_scale),
+    )
+
+
 def main() -> None:
     parser = argparse.ArgumentParser()
     parser.add_argument("--m", type=int, nargs="+", default=[1, 2, 4, 8, 32])
@@ -266,8 +298,8 @@ def main() -> None:
     tier1 = _prepared(4, materialized, 401, device)
     mixed_tier0 = mixed_tier1 = rotations = None
     if args.variant in ("both", "mixed"):
-        mixed_tier0 = _pad_projection_major_w13(tier0, K3_EXPERTS)
-        mixed_tier1 = _pad_projection_major_w13(tier1, K4_EXPERTS)
+        mixed_tier0 = _pad_exact_tier_storage(tier0, K3_EXPERTS)
+        mixed_tier1 = _pad_exact_tier_storage(tier1, K4_EXPERTS)
         rotations = _padded_rotations(tier0, tier1, device)
     global_map, descriptor = build_ordered_maps(K3_EXPERTS, K4_EXPERTS, device=device)
     map0 = torch.cat(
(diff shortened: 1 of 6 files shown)


malaiwah:
## Independent AIBeast field validation: exact r25 / #117 head — PASS

I independently qualified the released implementation on the full 3.36-bpw
checkpoint, including both of its real mixed partitions.

### Immutable inputs

- Image: `voipmonitor/vllm:gilded-gnosis-v20-vllmf5981f1-si978cdb3-fi801d57a-cu132-20260803-r25`
- Manifest: `sha256:042936fd8d9e4c2aa579ab9b736dd0a2faf2678c6ba36bf4dfce7db566c6fd11`
- PR head: `cfeee9b42d21c19a74d85ed5576f8387168df53c`
- Shipped `mixed_trellis.py` SHA-256: `edf3790e04252fb3d65507ff778e8af1d497dcd119ec0cd7dd61c3af4d47bc82`
- Checkpoint: `willfalco/GLM-5.2-EXL3-TR3-3.36bpw@8d9aa923a17502675ca23737349b67f2e66bb69d`
- Hardware: 4x RTX PRO 6000 Blackwell 96 GiB, driver 595.71.05, CUDA 13.2, 280 W/GPU; physical CUDA order `2,1,0,3`

### Appliance production shape

`TP4 / DCP4 / MTP3 probabilistic+standard / ONLINE_QUANT=exl3-b6 / dynamic-token NVFP4 MLA+FP8 RoPE / seq=8 / graph=32 / batch=3072 / EXL3 prefill capacity=3072 / GMU=.957 / fixed 2048 blocks = 524288 logical KV tokens`, plus LMCache 125 GiB DRAM and bounded 512 GiB NVMe.

This intentionally extends the release's 2,048-token/greedy gate with our
previously qualified PP-first scheduler and real agent speculation posture.

### Runtime and memory evidence

- layer 3 logged `tiers=((3, 206), (4, 50))`
- layers 4–77 logged `tiers=((3, 160), (4, 96))`
- both layouts planned `one-grid decode=32` and `one-grid prefill=3072/3072`
- model 81.07 GiB/rank; peak activation 3.29 GiB; non-Torch 0.44 GiB; graphs 0.21 GiB
- exact active GPU KV: 524,288 tokens
- no request failure, preemption, CUDA OOM, stale-layout symptom, or degeneration

### Matched benchmark

| metric | result |
|---|---:|
| cold unique-prefix PP 3K / 32K / 128K | 2362.8 / 2284.9 / 2143.8 tok/s |
| aggregate TG C1 / C2 / C4 / C8 | 100.7 / 162.2 / 240.1 / 297.0 tok/s |
| MTP MAL C1 / C2 / C4 / C8 | 3.13 / 3.43 / 3.41 / 3.02 |

Strict structured output with thinking passed. The release gate used an actual
521,275-token prompt and recovered 5/5 needles at 1/25/50/75/99% depth in
300.8 s, with no degeneration or OOM.

I also qualified the exact #117 source over the older r20 integration before
r25 arrived; its 3K/32K/128K PP was 2350.8/2276.4/2128.9 tok/s, C1 was 118.9
tok/s, and the same 521K five-depth gate passed. The r25 packaged result is
therefore consistent with the exact-source candidate rather than only passing
at the unit/kernel-test layer.

Conclusion: #117 fixes the actual 206/50 versus 160/96 geometry and is ready
for this four-GPU production workload from my side.

PR #135, opened 2026-08-09
fix(gemm): preserve dense API contracts with block FP8

voipmonitor:
## Summary

- preserve the existing `fused_quant_bf16` and FP6 `row_scale` API contracts when adding the regular K128 block-FP8 path
- keep block-FP8 accumulator coordinates in distinct CuTe DSL locals
- add a focused signature regression test for both legacy parameters

## Root cause

Commit `fe92b95` added `block_fp8`, but two signatures replaced existing parameters instead of appending the new one:

- `DenseGemmKernel.fused_quant_bf16`
- `dense_gemm.row_scale`

The implementation and callers still used those parameters. This caused model warmup to fail on FP6 row-scale paths and removed the explicit fused-quant compile-key input.

The same change also introduced a block-FP8 local named `c_identity` in the shared CuTe DSL kernel. Legacy fused-quant compilation uses that name in other compile-time branches. CuTe therefore observed `None` on one control-flow path and a tensor on another and rejected the first real DSpark request with `TYPE_UNSTABLE_JOIN`. Distinct `block_c_identity` / `block_coord_mn` locals remove that cross-family type join without changing either kernel's arithmetic.

## Validation

- Ruff and `git diff --check`: pass
- targeted API, FP6 row-scale, block-FP8 correctness and graph tests: 9 passed
- legacy fused-quant GPU oracle plus CUDA graph replay: passed
- DS4-Flash-0731 TP4/K5 full-model boot: target, DSpark and context-KV FULL CUDA graph capture passed
- 64 concurrent deterministic requests: 64/64 correct, no runtime errors
- matched remote-host performance versus r31:
  - C1: 238.7 vs 229.7 tok/s
  - C64: 2551.3 vs 2540.5 tok/s
  - prefill 8k/64k: 15187/14904 tok/s

The full-model test used GPUs 4-7 on the same direct-attached host as the r31 reference.

## Summary by CodeRabbit

* **New Features**
  * Added optional per-row scaling support for MX-FP6 dense matrix multiplication.
  * Added per-launch control over fused BF16 quantization.
  * Preserved the existing output rounding behavior for scaled operations.

* **Documentation**
  * Documented the new row-scaling behavior and optional controls.

* **Tests**
  * Added regression coverage to verify the public API defaults remain available.

### DIFF
--- a/b12x/_lib/dense_gemm.py
+++ b/b12x/_lib/dense_gemm.py
@@ -630,6 +630,7 @@ class DenseGemmKernel:
         mxfp6_fmt_b: Optional[str] = None,
         b_packed: bool = False,
         plain_fp8: bool = False,
+        fused_quant_bf16: Optional[bool] = None,
         block_fp8: bool = False,
     ):
         # When set, A/B operands are MX codes carried in Float8E4M3FN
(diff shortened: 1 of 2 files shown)


lukealonso:

PR #136, opened 2026-08-09
fix(gemm): restore capture-safe K6 small-M dispatch

voipmonitor:
## Summary

- Restore the established cooperative K6/MCG dense path for `M <= 128`.
- Keep the gate narrow: K6, MCG, unpaired payload, FP16 compute, and no external H128 callback. SQG/QSRT, paired payloads, BF16, and externally rotated calls remain on the generic scheduler.
- Reuse caller-owned output and rotation buffers, so vLLM decode graph capture does not require the generic W4A16 GEMM scratch arena.

## Root cause

`master` at `9bbae67` removed the K6/MCG small-M dispatch while retaining its implementation and CUDA-graph regression test. vLLM intentionally provisions only a one-element generic scratch placeholder for this qualified small-M path. Falling through to the generic scheduler therefore fails during graph capture with `W4A16 GEMM scratch is not initialized for CUDA graph capture`.

The fix restores the original ownership contract rather than increasing persistent scratch or disabling CUDA graphs.

## Validation

- Current `master`: `test_k6_small_m_cuda_graph_replay_is_stable` fails during capture.
- This branch: all K6 small-M tests -> **17 passed, 48 deselected** on SM120, including numerical comparison against the generic CuTe path and CUDA-graph replay.
- Ruff and `git diff --check`: pass.
- Live GLM-5.2 EXL3 3.36 bpw, TP4/DCP1/MTP3: FULL and PIECEWISE graph capture complete and coherent generation passes when combined with the mixed-ABI repair.
- Warm synthetic C1 regression gate: 139.5 and 144.4 tok/s.

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved quantized matrix multiplication tile selection for more efficient low-bit layouts.
  * Added a specialized path for small FP16 inputs to accelerate K6 MCG operations.
  * Reduced unnecessary metadata, transformations, and scheduling for eligible workloads.

### DIFF
--- a/b12x/moe/_shared/kernels/w4a16/kernel.py
+++ b/b12x/moe/_shared/kernels/w4a16/kernel.py
@@ -4199,12 +4199,10 @@ class W4A16GemmKernel:
             for jj in cutlass.range_constexpr(4):
                 local_n16 = Int32(4) * w_n + Int32(jj)
                 tile_base = Int32(0)
-                if cutlass.const_expr(int(low_bits) == int(high_bits)):
-                    tile_base = kt_base_u32 + local_n16 * Int32(8 * low_bits)
-                    wa[jj], wb[jj] = self._load_trellis256_pair_tile_windows(
-                        b_region, tile_base, lane, low_bits
-                    )
-                elif logical_k16 < Int32(8):
+                if (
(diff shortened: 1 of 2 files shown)


lukealonso:

PR #129, opened 2026-08-08
feat(moe): serve canonical Fruit QSRT atoms

malaiwah:
## Summary

- Serve canonical Fruit `qsrt_atoms_v1` geometry (`H=1024`, `I=512`, 96 fixed atoms per expert) through the public B12X MoE API.
- Decode per-expert P24/P33 pair modes and physical atom rotation without reconstructing dense expert weights.
- Preserve W4A16 prefill/reference execution and select W4A8 only for bounded decode rows.
- Support per-expert SiLU/SiTU metadata and fail closed on malformed layouts, unsupported activations, mixed gate/up modes, and invalid route maps.
- Keep steady-state storage caller-owned and CUDA-graph replay safe.

Current canonical head: `f4064d06f029240040a128388f32c861a535ad68`.

## Hot path

- FC1 input rotation and MXFP8 quantization use one fixed CUDA kernel rather than separate rotate/quantize launches.
- A route map is materialized once; FC1 and FC2 consume it directly instead of building Python-side old/new expert partitions and concatenating outputs.
- FC1 publishes one route-major output row, eliminating multipart accumulation for independently selected gate/up pair modes.
- FC2 remains route-major and preserves exact top-k weighting and physical-rotation semantics.
- W4A8 decode expands fixed-stride trellis payload directly into E4M3 MMA operands; W4A16 remains the graph-safe prefill/reference and unsupported-shape fallback.

## Verification

The current head is the exact B12X source fingerprinted into the final runtime image and sealed publication. The full 2,816-expert Fruit Instruct package loaded every routed and MTP layer under vLLM `FULL_AND_PIECEWISE` CUDA graphs. Runtime-path evidence records W4A16 prompt/prefill and W4A8 decode graph capture and replay, including the packaged MTP layer.

The sealed matched protocol used one immutable RTX 5090 image, launch order QSRT → BF16 → SIQ, TP1, `max_num_seqs=1`, identical prompt tokens/settings, and three warmed repetitions per arm:

| Arm | Median client-observed generated-token rate | Loader weight memory |
|---|---:|---:|
| BF16 | 445.04 tok/s | 9.730 GiB |
| SIQ | 428.63 tok/s | 3.240 GiB |
| QSRT | 425.67 tok/s | 3.100 GiB |

[...]

PR #137, opened 2026-08-09
fix(moe): realign mixed Trellis with QSRT ABI

voipmonitor:
## Summary

- Realign the existing mixed K3/K4 MCG launcher with the shared W4A16 ABI added by the fixed-payload QSRT work.
- Supply the new LUT and paired-M8 helper arguments at every direct call site, while keeping the established mixed checkpoint path explicitly on the MCG codebook.
- Restore qualification of the existing 512x32 ultra-wide FC2 tile used by the GLM-5.2 mixed prefill geometry.
- Add AST contract tests so future shared-kernel signature changes fail in unit tests instead of at CuTe compilation or model startup.

## Root cause

`master` at `9bbae67` changed several shared W4A16 signatures without updating the existing mixed K3/K4 direct calls. Arguments shifted position, so `active_m` received a callable and compilation failed with `functools.partial * Int32`. The forced GLM FC2 tile was also rejected by the generic tile floor despite having a valid e4m3-k32 geometry.

This PR is deliberately narrower than #129: it restores the already-supported MCG mixed path on current `master`; it does not adopt the broader Fruit/SQG serving feature. If #129 is rebased, it should preserve or supersede these ABI regression contracts explicitly.

## Validation

- Current `master`: `tests/moe/test_w4a16_mixed_trellis.py` -> **10 failed, 5 passed**.
- This branch: mixed suite plus new ABI contracts -> **18 passed** on SM120.
- Ruff and `git diff --check`: pass.
- Live GLM-5.2 EXL3 3.36 bpw, TP4/DCP1/MTP3: mixed tiers load, FULL and PIECEWISE graph capture complete, coherent generation passes.
- Warm synthetic C1 regression gate after the paired K6 fix: 139.5 and 144.4 tok/s; this is an MTP-friendly stability gate, not a general workload headline.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mixed-trellis processing for low-bit layouts and equal-bitrate tile pairs.
  * Enabled an additional FC2 specialization for supported K/32 scaling configurations.
  * Corrected codebook and dynamic pair handling for more reliable kernel execution.

* **Tests**
  * Added coverage to validate mixed-trellis dispatch, argument contracts, and internal kernel calls.

### DIFF
--- a/b12x/moe/_shared/kernels/w4a16/kernel.py
+++ b/b12x/moe/_shared/kernels/w4a16/kernel.py
@@ -441,7 +441,17 @@ def _candidate_tile_fits(
         or int(tile_k) % scale_group_size != 0
     ):
         return False
-    if int(tile_n) < 64 or int(tile_k) < 64 or int(cta_threads) < 128:
+    ultra_wide_fc2 = (
+        int(tile_n) == 512
+        and int(tile_k) == 32
+        and int(cta_threads) == 256
+        and _scale_group_size(scale_format) == 32
+    )
+    if (
+        int(tile_n) < 64
+        or (int(tile_k) < 64 and not ultra_wide_fc2)
+        or int(cta_threads) < 128
+    ):
         return False
     smem_bytes = _shared_memory_footprint(
         cta_m_blocks=cta_m_blocks,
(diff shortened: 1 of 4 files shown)


lukealonso:

PR #201, opened 2026-08-15
security: collectively agree hierarchical IPC contracts

### LINKED ISSUE #179: security: collectively agree hierarchical all-reduce IPC layout
## Summary

Hierarchical all-reduce ranks independently derive layouts/protocol used against peer IPC allocations.

Ranks are misconfigured or one admitted rank provides a divergent contract.

## Evidence

- [`b12x/comm/pcie/pcie_hierarchical.py:178-263`](https://github.com/local-inference-lab/b12x/blob/954fd0174c49502b62547a01f09c404029e6035a/b12x/comm/pcie/pcie_hierarchical.py#L178-L263) builds rank-local resources/mappings.
- Remote pointers are interpreted using local offsets/capacities, making disagreement memory-safety relevant.

## Impact

Out-of-bounds peer access, GPU failure, corrupted results, and deadlock.

## Recommended fix

- Collectively agree the full layout/protocol tuple before mapping.
- Abort all ranks coherently on mismatch.
- Test every size, offset, topology, and schedule field.

## Verification notes

malaiwah:
Closes #179.

Adds collective topology/ABI agreement before hierarchical all-reduce IPC allocation, coordinated construction rollback, and ordered lifecycle teardown.

Verification: 37 focused contract tests passed; 3 GPU-only tests skipped on the local host.

### DIFF
--- a/b12x/comm/pcie/pcie_allreduce.py
+++ b/b12x/comm/pcie/pcie_allreduce.py
@@ -84,8 +84,9 @@ class PCIeAllReduce:
                 single_channel=single_channel,
             )
         else:
-            if max_size < torch.bfloat16.itemsize:
-                raise ValueError("max_size must hold at least one BF16 element")
+            # max_size validation is handled by the constructor's collective
+            # preflight so that one rank's invalid argument does not strand
+            # peers in the contract exchange.
             runtime = PCIeHierarchicalAllReduce(
                 exchange_group=exchange_group,
                 device=device,
@@ -142,10 +143,14 @@ class PCIeAllReduce:
                 raise ValueError(
                     "peer_input_ptrs are unavailable for hierarchical all-reduce"
                 )
+            if blocks is not None:
+                raise ValueError(
+                    "blocks override is not supported for hierarchical "
+                    "all-reduce; the schedule is frozen in the catalog"
+                )
             return self._runtime.all_reduce(
                 inp,
                 out=out,
-                blocks=blocks,
                 stream=stream,
             )
         if blocks is not None:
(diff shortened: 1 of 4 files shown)


malaiwah:
Hardware: 2× NVIDIA RTX PRO 6000 Blackwell Server Edition (97,887 MiB each), driver 595.58.03, PCIe SYS topology, peer access enabled. Environment: torch 2.13.0+cu130, CUDA 13.0, CUTLASS DSL 4.6.0. CUDA_VISIBLE_DEVICES=0,1.

GPU evidence at `7ed1cfe7b4497aaf880d5c72eed337c903b3ca47`:

```
python -m pytest -q -rs --durations=20 tests/comm/test_pcie_hierarchical_contract.py tests/comm/test_pcie_hierarchical_gpu.py
```

Result: **37 passed, 3 skipped in 1.25s** (wall interval 1.93s). Negative mismatch, malformed-handle, rollback, close-order, and simulated 12-participant matching cases passed. The three real GPU tests explicitly require `B12X_RUN_PCIE_HIERARCHICAL_TEST=1` and TP12; they were not executed because the provisioned runner has 2 GPUs and current provider availability is at most 8. Therefore this PR still lacks real TP12 positive GPU evidence and should not be called GPU-qualified.

lukealonso (changes requested):
This breaks existing hierarchical IPC behavior in two concrete ways. The default buffer catalog contains only the maximum size while lookup accepts only exact sizes, so normal `size <= max` requests no longer resolve. It also removes runtime capture/registration methods that the public wrapper still calls. The GPU tests that would exercise these paths are skipped here.

Please restore the public wrapper contract and size-selection semantics, then test real multi-rank setup, eager all-reduce, CUDA-graph capture/replay, and teardown on the target topology. This should not merge until those functional regressions are fixed.

lukealonso:
Closing rather than merging. The threat model is rank misconfiguration or an already-compromised ProcessGroup member, not a new security boundary. The patch adds roughly 1,000 production lines of hierarchical IPC agreement, rollback, and teardown machinery, with GPU qualification skipped. This is distributed reliability hardening with substantial lifecycle complexity, not a surgical security fix.

PR #211, opened 2026-08-15
security: reject out-of-range MSA q2k blocks

### LINKED ISSUE #159: security: reject out-of-range MSA q2k block indices
## Summary

The MSA block-sparse metadata path does not fully bound positive q2k block selections before cache-region selection.

A caller supplies or graph-mutates block-sparse selection metadata.

## Evidence

- [`b12x/attention/paged/_scratch.py:1392-1439`](https://github.com/local-inference-lab/b12x/blob/954fd0174c49502b62547a01f09c404029e6035a/b12x/attention/paged/_scratch.py#L1392-L1439) prepares MSA block metadata without a complete positive upper-bound check.
- Downstream kernels consume selected blocks for paged addressing; shape checks do not validate each mutable value.

## Impact

Out-of-bounds device reads, CUDA-context failure, service denial of service, and possible disclosure through output.

## Recommended fix

- Validate every active q2k block against logical and physical ranges.
- Guard graph-replayed values in kernel.
- Test sentinel, exact upper bound, and large positive values.

## Verification notes

malaiwah:
Closes #159.

Bounds graph-live q2k block selections against logical and physical paged-cache regions before address formation across paged and extend routes.

Verification: 14 focused SM121 GPU tests passed; py_compile/Ruff/diff checks passed.

### DIFF
--- a/b12x/attention/paged/_forward.py
+++ b/b12x/attention/paged/_forward.py
@@ -1084,6 +1084,8 @@ def paged_attention_forward(
             union_blocks=msa_union_blocks,
             union_masks=msa_union_masks,
             union_counts=msa_union_counts,
+            page_table_width=int(page_table.shape[1]),
+            page_size=int(plan.page_size),
         )
 
     q_arg = _to_kernel_tensor(q, _torch_to_cutlass_dtype(q.dtype))
--- a/b12x/attention/paged/_scratch.py
+++ b/b12x/attention/paged/_scratch.py
@@ -429,6 +429,27 @@ def _make_plan_metadata_cache(
     )
 
 
+
+
+_MSA_BLOCK_TOKENS = 128
+_MSA_TOPK = 16
+
+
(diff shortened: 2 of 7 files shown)


lukealonso (changes requested):
The MSA bounds defense changes hot-kernel branching and 64-bit index arithmetic and also changes launch-grid behavior. These changes can alter CTA count, live ranges, register use, and occupancy even when all q2k blocks are valid.

Please add adversarial large-index correctness coverage and report target-GPU MSA decode/prefill plus CUDA-graph timings on the exact head. Include SASS/resource deltas, launch geometry, and raw samples so the safety checks can be evaluated independently from the grid change.

lukealonso:
Closing rather than merging. The implementation changes paged-attention forward, scratch, and device-kernel address/masking paths. That directly violates the zero-kernel-code-change constraint. The issue remains open because graph-mutated q2k values cannot be fully bounded by a one-time host check.

PR #212, opened 2026-08-15
security: bound sparse and compressed MLA cache indices

### LINKED ISSUE #158: security: bound sparse/compressed MLA live cache indices
## Summary

Sparse and compressed MLA allow runtime live lengths/selected cache indices to reach shared gather addressing without complete positive upper-bound validation.

A caller or serving integration controls selection metadata or cache lengths.

## Evidence

- [`b12x/attention/_shared/mla/api.py:500-575`](https://github.com/local-inference-lab/b12x/blob/954fd0174c49502b62547a01f09c404029e6035a/b12x/attention/_shared/mla/api.py#L500-L575) validates structure but does not establish all live index bounds against physical capacity.
- Sparse/compressed decode and extend paths feed the metadata into shared issue/gather address calculation.
- Handling negative sentinels does not bound high positive indices.

## Impact

Out-of-bounds GPU reads, context failure, denial of service, and possible disclosure in returned values.

## Recommended fix

- Tie bindings to physical extents and check every live positive index.
- Retain kernel predicates for graph replay.
- Test `capacity-1`, `capacity`, and high recycled IDs.

## Verification notes

malaiwah:
Closes #158.

Binds sparse/compressed MLA selection metadata to live logical and physical cache extents and carries replay-time bounds into shared gather addressing. Invalid active indices are rejected or masked before forming cache addresses.

Verification: 23 focused SM121 GPU tests passed; py_compile/Ruff/diff checks passed.

### DIFF
--- a/b12x/attention/_shared/mla/api.py
+++ b/b12x/attention/_shared/mla/api.py
@@ -119,6 +119,101 @@ def _validate_tensor_storage_bounds(tensor: torch.Tensor, *, name: str) -> None:
         )
 
 
+def _sparse_mla_slot_capacity(
+    kv_cache: torch.Tensor, page_size: int
+) -> int:
+    """Physical token-slot capacity of a sparse/compressed MLA KV cache.
+
+    Rank-3 [num_pages, page_size, record_bytes] and rank-2 [num_pages,
+    page_bytes] layouts are both supported.  The capacity is the total number
+    of addressable token slots (num_pages * page_size); the in-kernel gather
+    clamps any live index >= capacity to a safe sentinel.
+    """
+    page_size = int(page_size)
+    if kv_cache.ndim == 3:
+        num_pages = int(kv_cache.shape[0])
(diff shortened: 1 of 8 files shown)


lukealonso (changes requested):
The eager validation path calls GPU `.item()` and `.max().item()` per row, forcing repeated device-to-host synchronization and making latency scale with row count. In addition, two prefill MG call sites no longer match the callee signature: one omits required positional structure and another passes duplicate extras.

Please fix every call-site/callee contract first, replace host-synchronous row validation with a graph-safe device status path or prevalidated metadata, and add direct tests for both prefill MG routes. Provide target-GPU sparse/compressed MLA decode, prefill, and CUDA-graph measurements before reconsideration.

lukealonso:
Closing rather than merging. The implementation changes shared sparse/compressed MLA kernel ABI, gather addressing, masking, and launch plumbing. That directly violates the zero-kernel-code-change constraint and is not a host-only boundary fix. The issue remains open.

PR #189, opened 2026-08-13
security: enforce dense current-stream contract

malaiwah:
## Problem
The dense GEMM APIs accept a caller-selected CUDA stream, but the complete composite operation includes PyTorch allocation, fill, reduction, and epilogue work around raw CuTe launches. A non-current explicit stream therefore needs both execution dependencies and allocator-lifetime ownership at one boundary.

## Implementation
- Preserve the existing `stream=None` path. It enters the original implementation after one branch and performs no CUDA stream lookup, synchronization, or `record_stream` work.
- Run the complete dense, fused-quant, or grouped fused-quant operation under the selected CUDA device and stream.
- For a distinct explicit stream, make the launch stream wait for the caller's current stream, record each unique external storage once, and make the caller stream wait for completion before returning.
- Avoid allocator recording and stream waits when the explicit stream is already current.
- Accept Torch streams and raw handles, validating raw-handle device ownership before wrapping them.
- Keep custom-op and raw-launch layers free of duplicate stream validation or ownership bookkeeping.

## Cost model
- `stream=None`: one required Python branch; no stream resolution, events, or allocator bookkeeping.
- Explicit current stream: stream/device resolution and a device context; no events or allocator bookkeeping.
- Explicit non-current stream: two stream dependencies and one deduplicated `record_stream` pass over caller-owned storage.

## Validation
- CPU/source contract tests on head `7b7420662f32871e8890edec78ec0c16dd40ca7d`: 6 passed.
- Python compilation, test collection (8 tests), and `git diff --check`: passed.
- Two target-GPU regressions cover producer/launch/consumer ordering with allocator churn and explicit capture-stream replay. They remain pending on this exact head because all shared GPUs are occupied.
- Representative dense GEMM, weight-only, and block-FP8 eager/CUDA-graph measurements are required before merge.

Closes #153

PR #188, opened 2026-08-13
security: validate FP6 dense artifacts before launch

malaiwah:
## Problem

The standalone FP6 dense loader trusted safetensors metadata independently from packed tensor geometry. A malformed artifact could declare output dimensions inconsistent with its packed rows and scale storage, then reach raw CUDA launch geometry.

## Fix

- validate the exact `b12x_fp6_dense_weight_v1` tensor and metadata schema on CPU before tensor materialization
- require packed rows/columns, padded scale bytes, scalar scale, formats, contiguity, and device coherence to match producer contracts
- reject inconsistent raw expanded-weight output geometry before launch
- preserve valid non-128-aligned output widths; only K must satisfy the 32-element FP6 scale group

## Verification

- Linux/CUDA: `pytest tests/quantization/test_fp6_dense_artifact_validation.py -q` — 34 passed
- `ruff check` on source and focused tests
- `python -m py_compile` on source and focused tests
- independent adversarial peer review approved the implementation and follow-up schema/materialization-order tests; current head adds only its requested docstring correction

Closes #152

PR #213, opened 2026-08-15
security: validate serialized FP6 dense weights

### LINKED ISSUE #163: security: enforce full RHS and output contracts in dense_gemm
## Summary

The generic `dense_gemm` forwards raw pointers without fully checking RHS K/L compatibility or caller output shape, dtype, device, capacity, layout, and overlap.

Public GEMM/projection wrappers or artifact loaders forward malformed tensors; malicious FP6 metadata is one concrete route.

## Evidence

- [`b12x/_lib/dense_gemm.py:7163-7169`](https://github.com/local-inference-lab/b12x/blob/954fd0174c49502b62547a01f09c404029e6035a/b12x/_lib/dense_gemm.py#L7163-L7169) derives launch geometry without a complete cross-operand contract.
- [`b12x/_lib/dense_gemm.py:7497-7515`](https://github.com/local-inference-lab/b12x/blob/954fd0174c49502b62547a01f09c404029e6035a/b12x/_lib/dense_gemm.py#L7497-L7515) forwards optional `out` as the destination.
- Kernel geometry, not destination capacity, determines write extent.

## Impact

OOB reads/writes, wrong-device access, alias races, stale-data disclosure, and CUDA-context failure.

## Recommended fix

- Validate K and batch/L agreement.
- Validate exact output shape/dtype/device/layout/alignment and backing bytes.
- Reject unsafe aliases.
- Add direct low-level boundary tests.

## Verification notes

malaiwah:
Closes #163.

Validates serialized FP6DenseWeight metadata against packed values, swizzled scale storage, global scale, formats, and device placement when the prepared weight is constructed. Malformed safetensors metadata is rejected before any expansion, compilation, or launch.

The generic dense_gemm implementation is unchanged. Public linear wrappers already construct its operands; duplicating those contracts on every token would add hot-path Python work without creating a meaningful security boundary.

Verification:
- 31 focused FP6 construction and compatibility tests passed.
- Ruff, compileall, and diff checks passed.
- Cumulative PR diff contains no kernel-code changes.


lukealonso (changes requested):
The dense GEMM entry point now performs exhaustive Python storage, device, shape, stride, and pairwise overlap validation on every call. The overlap checks are quadratic in the number of buffers and sit directly on the M=1 eager dispatch path; stable prepared-weight invariants should not be recomputed per token.

Please move stable checks to prepared objects or a single admission boundary, retain only mutable launch invariants, and benchmark M=1 plus representative eager and CUDA-graph shapes. Rebase and reconcile this with #189/#194 so ownership, capture admission, and buffer validation form one measured contract.

lukealonso:
Closing as redundant with merged PR #188. The concrete artifact-loader boundary already validates FP6 metadata against packed values, scale storage, formats, and device placement during FP6DenseWeight construction. The remaining claim depends on malicious same-process calls to the private dense_gemm implementation; adding per-launch validation there would add hot-path overhead without establishing a meaningful security boundary. No kernel-code change is warranted.

PR #214, opened 2026-08-15
[II] Ignore inactive routes in dynamic MoE kernels

voipmonitor:
## Status

Implemented at `321c24a`. Targeted GPU qualification passed; the full repository
test suite was not rerun.

## Resulting behavior

The dynamic MoE route contract treats every expert ID outside
`[0, num_experts)` as inactive.

- Grouped routing excludes inactive entries from expert histograms, compacted
  storage, and task publication.
- Direct routing retains its graph-stable arithmetic task domain, marks inactive
  slots with zero valid rows, and skips them before loading expert weights or
  scales.
- The materialized M=1 producer derives one warp-uniform active-route signal and
  skips activation loads, quantization, packed writes, and scale writes when no
  route is active.
- Shared-input producers derive one token-level active-route signal after route
  allocation and skip activation loads, quantization, packed writes, fences,
  counters, and task publication when the token has no active route.
- Route-private producers use the existing warp-uniform route-valid signal to
  guard the expert scale load and the complete activation producer body.
- Tokens with no active routes retain the kernel's zero-initialized output.
- CUDA graph replay may change between valid, partially padded, and fully padded
  routing without rebinding or allocating workspace.

## Technical reason

vLLM's skip-padding MoE path uses expert ID `-1` for zero-weight padding routes;
the option defaults to enabled after
[vLLM #48979](https://github.com/vllm-project/vllm/pull/48979). Directly using
that ID in B12X expert histograms or prepared-weight addressing violates the
route storage bounds.

[local-inference-lab/vllm #291](https://github.com/local-inference-lab/vllm/pull/291)
maps `-1` to expert zero at the integration boundary. That mapping avoids the
invalid address but converts padding into executable expert-zero work. Defining
inactive routes in B12X preserves the skip-padding optimization and keeps the
contract valid for every B12X caller.

## Efficiency evidence

The inactive checks surround producer work rather than only predicating its
stores.

- A zero-initialized packed-activation probe on `83e306f` produced 256 nonzero
  elements for the fully inactive M=1 case and 768 for the fully inactive
  shared-input case. The same probes on `321c24a` produced zero packed elements
  and exact-zero outputs in both cases.
- `nvdisasm` on the `321c24a` W4A8 objects shows a uniform M=1 branch at `0x840`
  jumping to `0x1250`, past the activation/quantization/store body, and a
  shared-input branch at `0x1c30` jumping to `0x3150`, past both cooperating
  warp partitions.
- A valid-route production-path diagnostic at
  `M40/E64/H1024/I256/top-k6`, warm CUDA graph replay, used five samples of
  100,000 replays per head. `83e306f` measured `32.770 us`; `321c24a` measured
  `32.786 us` (`+0.047%`). Allocated memory remained `62,860,288` bytes across
  replay in both arms. This Max-Q diagnostic indicates no material active-route
  regression; it is not formal release qualification.

## Compatibility

Valid expert IDs retain the same routing, quantization, task geometry, and
output math. Active M=1 and shared-input producers add a short lane-zero route
scan plus a warp shuffle. The implementation adds no allocation, workspace, or
binding changes. Negative and positive out-of-range IDs are both inactive.

## Validation

Source: `321c24a` on B12X `master` base `d3fc4bd`.

GPU: NVIDIA RTX PRO 6000 Blackwell Max-Q Workstation Edition,
`GPU-ac6fcbb2-ae5f-231d-cc3e-e843c305baff`; Torch `2.12.0+cu130`, CUDA `13.0`,
CUTLASS DSL `4.6.0`.

- `7 passed`: materialized M=1, shared-input materialization, grouped W4A8,
  production NVFP4 inactive routing, graph mutation/replay, and both dynamic
  prefill corpus cases.
- The existing `M=128` production dynamic-prefill corpus remains present; the
  `M=40` inactive-route corpus is a separate test.
- `py_compile`, `ruff check --ignore F401`, and `git diff --check`: passed.

## Merge recommendation

Ready to merge based on targeted GPU qualification. The full repository suite was not run.

### DIFF
--- a/b12x/moe/_shared/kernels/dynamic.py
+++ b/b12x/moe/_shared/kernels/dynamic.py
@@ -2998,7 +2998,13 @@ class MoEDynamicKernelBackend:
             hist_idx = flat_tid
             while hist_idx < total_pairs:
                 expert_id = topk_ids[hist_idx].to(Int32)
-                atomic_add_global_i32(get_ptr_as_int64(row_counts, expert_id), Int32(1))
+                # A route outside the local expert domain is inactive. vLLM
+                # uses -1 for scheduler padding, and expert-parallel callers
+                # may use the same contract for non-local routes.
+                if expert_id >= Int32(0) and expert_id < num_experts:
+                    atomic_add_global_i32(
+                        get_ptr_as_int64(row_counts, expert_id), Int32(1)
+                    )
                 hist_idx += flat_stride
 
             self._resident_grid_barrier(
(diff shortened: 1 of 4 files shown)


voipmonitor:
## Qualification evidence

**Status: qualified.** Commit `321c24a7ef60174cd6131d932f43bb84a4f3a60f` was validated in the published Infernal Invocation r15 image:

```text
voipmonitor/vllm:infernal-invocation-vllm068fc8e-b12x96e5d3d-fi1ac6942-cu133-torch213-20260815-r15
registry digest: sha256:f1b13c8604b274212e1164def7d4ed7a4cac9e4f7fa06fa1739730195eca4e18
```

The qualified contract is:

- route identifiers below zero or outside the expert range are inactive;
- inactive routes do not contribute to histograms, compact route storage, weight access, quantization, or route fanout;
- valid routes retain their arithmetic and output layout;
- `VLLM_MOE_SKIP_PADDING=0` remains a compatibility override.

Validation:

- 29 physical-GPU B12X tests passed across dynamic Trellis, W4A8 dynamic, and standard MoE corpora;
- target-only TP2/DCP1 captured FULL CUDA graphs and sustained 125.7 tok/s at CC1;
- fixed probabilistic DSpark K5 captured FULL target and draft CUDA graphs and sustained 191.3 aggregate tok/s at CC1;
- the arithmetic content probe returned exactly `{"a":42,"b":126}`;
- runtime logs contained zero post-startup errors;
- an isolated A/B composition reached target-only C1/C4/C8 of 125.8/339.9/481.1 tok/s and 8K/64K prefill of 12,646/13,195 tok/s.

vLLM PR #291 is excluded from the image because mapping inactive routes to expert zero performs unnecessary expert-zero work.

Evidence:

- https://github.com/local-inference-lab/blackwell-llm-docker/blob/main/validation/infernal-invocation-r15-remote-gpu.json
- https://github.com/local-inference-lab/blackwell-llm-docker/blob/main/validation/infernal-invocation-r15/runtime-summary.json
- https://github.com/local-inference-lab/rtx6kpro/blob/master/models/ds4dspark-infernal-invocation-r15.md

PR #215, opened 2026-08-15
Preserve unaligned multi-row FP8 linear outputs

voipmonitor:
## Status

**Implemented and GPU-qualified.** The source implementation is B12X revision
`8cb3f16adc6499fd8d87b82a42348490973e10db`; revision
`b3c2c162064ea9fab118f08922577aca5912e7e6` adds the reproducible qualification
receipt without changing executable source.

## Behavior

`dense_gemm` writes complete multi-row BF16 or FP16 outputs when the logical row
stride is not 16-byte aligned. When the caller leaves the MMA tile and operand
order unspecified, a single-group MXFP8 operation adapts the tuned default plan
to supported swapped output storage. The policy preserves the default planner's
dependence on M, N, K, `expected_m`, and SM count.

For the Kimi-K3 tensor-parallel projection shape `M=3, N=132, K=7168`, the
policy selects the swapped TMA `(64, 32)` MMA tile. Swapped MXFP8 storage uses
BK128 because BK64 packed-scale staging requires an unswapped 128-row weight
operand.

Explicit unswapped plans with an unaligned row stride fail before launch.
Grouped `L > 1` operations cannot use swapped output storage and receive a
diagnostic that requires N padding or another supported storage path.

## Technical reason

The unswapped TMA epilogue requires each outer output stride to be 16-byte
aligned. A 132-column BF16 output has a 264-byte row stride, so an unswapped
multi-row launch can leave rows unwritten. Operand swapping makes the logical
output row the contiguous TMA dimension without padding the public tensor.

A fixed swapped `(64, 64)` policy is invalid for some supported shapes because
it discards the tuned tile selection and can combine swapped operands with an
incompatible BK64 scale-staging plan. The implementation instead transforms the
selected plan and forces BK128 only when swapped MXFP8 storage is required.

## Compatibility

- Aligned outputs and single-row outputs retain default dispatch.
- Caller-specified valid plans retain their requested tile and operand order.
- MXFP8 and tensor-FP8 wrappers delegate narrow-output policy to `dense_gemm`.
- CUDA Graph capture and replay are supported.
- No model-specific branch or tensor-parallel-size check is introduced.

## Validation

Conditions: NVIDIA RTX PRO 6000 Blackwell Workstation Edition
`GPU-d8438b2d-f000-a617-5dcc-0197ce0365a3`, driver 610.57.04, CUDA 13.3,
PyTorch 2.13.0, and container
`voipmonitor/vllm:kimi-k3-tp16-vllm2ddc210-b12x3bce5d8-cu133-torch213-20260816-r1`
with source revision `8cb3f16adc6499fd8d87b82a42348490973e10db` mounted at
`/src/b12x`.

- 50 targeted tests passed across
  `tests/gemm/test_dense_gemm_expected_m.py`,
  `tests/gemm/test_mxfp8_linear.py`, and
  `tests/gemm/test_tensor_fp8_linear.py`.
- Ruff lint, Ruff format verification, and Git whitespace validation pass for
  every changed Python file.
- The public tensor-FP8 regression shape
  `M=2, N=16386, K=1024, expected_m=2048` passed reference comparison under
  CUDA Graph replay with the supported swapped BK128 plan.
- Grouped validation covers `M=1, L=2, N=132` and confirms that both an
  unaligned unswapped plan and an explicitly swapped grouped plan fail with the
  supported padding diagnostic.
- The complete `mxfp8_linear.mm` operation, including BF16 activation
  quantization, was measured under CUDA Graph replay with alternating AB/BA
  order for 1,000 sample pairs. Swapped `(64, 32)` measured 18.080 us median;
  swapped `(64, 64)` measured 22.176 us median. Both matched the dequantized
  reference with zero mismatches. The selected `(64, 32)` plan used 81.53% of
  the `(64, 64)` median time.
- The receipt binds each plan to the hash-verified CUTLASS objects created from
  an empty compile cache and to CUDA 13.3 PTXAS V13.3.73.
- Reproducible evidence and raw samples are stored in
  `validation/performance/mxfp8_linear_output_plan_sm120.md` and
  `validation/performance/mxfp8_linear_output_plan_sm120.json.gz`. The compressed
  receipt SHA-256 is
  `5d6532a9df5228e2f0fae3e964011158736f2078035fb35222ec8f6061e3fed2`.

Conclusion: unaligned single-group FP8 output uses a supported, graph-safe plan
that preserves the tuned shape policy, while unsupported explicit and grouped
storage configurations fail before CUDA launch.

## AI assistance

Implementation and validation were performed with AI assistance. The committed
tests and qualification receipt define the supported behavior.


lukealonso (changes requested):
The swapped direct-store mechanism is appropriate: it avoids the unaligned TMA epilogue instead of adding a predicate around work that still executes. However, the plan selection is not safe or performance-qualified as a general output-layout policy.

1. `dense_gemm_output_storage_plan()` returns `mma_tiler_mn=(64, 64), swap_ab=True` for every multi-row unaligned output, replacing the existing planner dependence on `M`, `N`, `K`, `expected_m`, and SM count. This produces a supported public-wrapper input that fails before launch: `M=2, N=16386, K=1024, expected_m=2048, sm_count=188` selects BK64, after which `_validate_mxfp8_bk64_plan` rejects the forced plan with `MXFP8 BK64 packed-scale staging requires an unswapped 128-row tile, got tile=(64, 64), swap_ab=True`. Adapt the existing tuned plan into a valid swapped plan, explicitly choose BK128 for this storage path, or constrain the advertised contract. Add a regression test for this interaction.

2. For the qualified `M=3, N=132, K=7168` shape, the reported proxy gives 0.237760 ms for swapped 64x32 and 0.247648 ms for the selected swapped 64x64 plan. The selected plan is 4.16% slower in the only attached comparison. Please run the complete linear wrapper and CUDA Graph path with balanced plan ordering, then select the supported winner or provide contrary evidence.

3. The new dense-GEMM error recommends `swap_ab=True` for `L > 1`, but `DenseGemmKernel.can_implement` rejects every swapped plan when `L != 1`. The diagnostic should recommend swapping only for `L == 1`; grouped outputs must be padded or handled by another supported storage path.

The aligned and single-row behavior is otherwise appropriately preserved. The focused CPU planner and guard suite passes locally (26 tests); the reported GPU qualification was not independently rerun.

Please also replace the `[II]` title prefix with a durable semantic title.

voipmonitor:
@lukealonso The requested plan-safety and qualification conditions are implemented at `47a1de543607b8c2e8055fb7ea5a37b96f15f07c`:

1. Unaligned output storage adapts the M/N/K/`expected_m`/SM-selected default tile instead of replacing it with a fixed `(64,64)` plan. Swapped MXFP8 storage forces BK128. The public `M=2, N=16386, K=1024, expected_m=2048` tensor-FP8 operation now passes reference comparison under CUDA Graph replay.
2. The complete `mxfp8_linear.mm` path, including activation quantization, was measured under CUDA Graph replay with alternating AB/BA order for 1,000 sample pairs. Swapped `(64,32)` is selected: 18.080 us median versus 22.176 us for swapped `(64,64)`, with zero reference mismatches in both arms. The raw receipt and reproduction command are committed under `validation/performance/`.
3. Grouped `L > 1` output cannot select swapped storage. Invalid grouped configurations now require padding N or another supported storage path; tests cover both unswapped unaligned output and an explicit grouped swap.

The PR is rebased onto `e68f812`, contains two commits (implementation and qualification evidence), has a semantic title without a branch prefix, and passes 50 targeted tests on an RTX PRO 6000 Blackwell GPU. Please re-review this head.

voipmonitor:
@lukealonso The reviewable head is `e8836543b772cf7db3b60aa64df96077cb2ee106` with two commits: source behavior at `bd424b7cf51b8096880c4259e2d4772b2d51e02b` and its qualification receipt.

The FP8 wrappers now delegate narrow-output policy to `dense_gemm`; no duplicate wrapper policy remains. The benchmark resolves the logical CUDA device to its physical PCI identity, requires an empty B12X compile cache, binds each plan to hash-verified CUTLASS objects and PTXAS, records first-replay and 1,000 warm samples, and explicitly identifies the measurement as a within-revision plan comparison. All qualification checks passed in P1 with throttle mask zero. The 50 targeted GPU tests also pass on this source revision.

PR #227, opened 2026-08-17
[II] Honor inactive routes in native W4A16 microkernels

voipmonitor:
## Behavior

Native ModelOpt W4A16 direct MoE execution treats an expert identifier outside `[0, resident_expert_count)` as an inactive route. The implementation never addresses weights or scales through an inactive identifier, and the inactive route contributes exactly zero. Valid routes retain their identifiers and weights. Caller-owned route tensors remain unchanged in eager execution and CUDA Graph replay.

Fixed-M fused FC1+FC2 launches sanitize their compile-time-bounded route table in shared memory. FC2-only launches accept runtime M and validate each route inline against the runtime resident-expert count. Route validity is evaluated at the source tensor width; an `int64` identifier is narrowed only after it passes the range check.

## Technical reason

`MoEMicroKernelW4A16SmallMDirect` bypasses the dynamic route scheduler. Direct weight and scale indexing with vLLM's `-1` padding sentinel can therefore read outside expert storage. FC2-only compilation uses a capacity bucket that may be smaller than runtime M, so FC2 cannot safely index a fixed-size staged route table. Narrowing an `int64` identifier before validation can also map `2**32` to expert zero.

The FC2 launch ABI carries the runtime expert count and performs the bounds check for every runtime route. Fixed-M fused execution stages sanitized identifiers and effective weights once per CTA. Invalid identifiers use expert zero only as an addressable placeholder with an exact-zero effective weight.

## Compatibility

- Scope: native direct W4A16 execution.
- Dynamic MoE, route-packed W4A16, and Trellis execution retain their behavior.
- Fixed CUDA Graph task geometry is unchanged; inactive arithmetic slots are address-safe and zero-contributing but are not compacted.
- The fixed-M shared-memory cost is `m * top_k * 8` bytes. Kimi-K3 decode at `M=1, top_k=16` uses 128 bytes per CTA.
- FC2-only runtime-M execution does not allocate a fixed route table.
- The contract is independent of tensor-parallel world size for every topology that selects the direct kernel.

## Validation

Environment: NVIDIA RTX PRO 6000 Blackwell, CUDA 13.3, PyTorch 2.13.0, CUTLASS DSL 4.6.2.

- Nine source-width tests passed for fused and FC2-only execution.
- Fused tests cover `int32` route tables at M=1/2/4/8 and `int64` at M=1.
- FC2-only tests cover `int32` and `int64` at M=3 with 256- and 512-element intermediate widths.
- Both `int64` paths use `2**32` as an inactive identifier.
- Assertions cover direct dispatch, eager execution, CUDA Graph replay, exact-zero inactive output, finite and nonzero valid output, oracle parity, and immutable route tensors.
- Compute Sanitizer reported `ERROR SUMMARY: 0 errors` for the runtime-M invalid-route cases.
- Generic NVFP4 compilation tests cover the fused body and both split phases through the real CuTe compiler.
- A composed B12X source with mixed Trellis support produced 329 passes and 18 skips.

Full-model qualification used official Kimi-K3 MXFP4, the Inferact seven-token DSpark draft, TP16/DCP16, FP8 target KV cache, native vision, LMCache, a 1,000,000-token model limit, and CUDA Graph capture size eight. The profile loaded 1,033,126 physical target KV tokens and completed graph capture. Median target-cycle rate was 31.519620 cycles/s versus 31.429716 cycles/s for the comparison image, a higher-is-better ratio of 1.0029.

The durable commands, source identities, raw samples, correctness results, and full-model conditions are recorded in [`validation/performance/w4a16_inactive_routes_sm120.md`](https://github.com/local-inference-lab/b12x/blob/0eba6ae99e0d1fad6ec268d8c291f498ec1dd4d9/validation/performance/w4a16_inactive_routes_sm120.md).

Status: **qualified** for native ModelOpt W4A16 fixed-M fused execution and FC2-only runtime-M execution on SM120.

## AI assistance

AI assistance was used to implement, review, test, benchmark, and document the change. The submitting human must review every changed line and be able to defend the behavior before merge.

### DIFF
--- a/b12x/moe/_shared/kernels/micro.py
+++ b/b12x/moe/_shared/kernels/micro.py
@@ -406,6 +406,7 @@ class MoEMicroKernelBackend:
         weight_layout: str = "modelopt",
         trellis_bits: int | None = None,
         trellis_coupled: bool = False,
+        stage_inactive_routes: bool = False,
     ):
         activation = normalize_moe_activation(activation)
         if int(compile_time_phase) not in {0, 1, 2}:
@@ -473,8 +474,10 @@ class MoEMicroKernelBackend:
         self.weight_layout_trellis256 = weight_layout == "trellis_t256"
         self.trellis_bits = 0 if trellis_bits is None else int(trellis_bits)
         self.trellis_coupled = bool(trellis_coupled)
+        self.stage_inactive_routes = bool(stage_inactive_routes)
         self.trellis_ksplit = 1
         self.trellis_scratch_u32 = 0
+        self.route_slots = 0
         self._cfg = None
         self.m_const = 0
         self.m1_fc2_onepass = False
@@ -497,6 +500,7 @@ class MoEMicroKernelBackend:
             self.weight_layout,
             self.trellis_bits,
             self.trellis_coupled,
+            self.stage_inactive_routes,
             self.trellis_ksplit,
             self.scale_format,
             self.e8m0_scale_layout,
@@ -511,6 +515,7 @@ class MoEMicroKernelBackend:
             self.m1_fc2_onepass,
             self.m1_fc2_rows_per_cta,
             self.launch_block_dim,
+            self.route_slots,
         )
 
     @cute.jit
@@ -894,6 +899,7 @@ class MoEMicroKernelBackend:
         self.m1_fc2_rows_per_cta = m1_fc2_rows
         self.launch_block_dim = _K_PER_CTA * 16 if m1_half_cta_fc2 else _BLOCK_DIM
         self.grid_x = grid_x
+        self.route_slots = m * cfg.num_topk
         # Required inter_fp32 element count: the per-token intermediate plus
         # the trellis K-split scratch tail. Callers must zero-initialize; the
         # kernel restores the tail to zero after each use.
@@ -2399,6 +2405,37 @@ class MoEMicroKernelBackend:
         bidx_x, _, _ = cute.arch.block_idx()
         tidx, _, _ = cute.arch.thread_idx()
         gdim_x, _, _ = cute.arch.grid_dim()
+        if cutlass.const_expr(self.stage_inactive_routes):
+            # The native direct path bypasses dynamic route packing, so stage
+            # its bounded route table once per CTA. Invalid IDs receive an
+            # addressable placeholder and an exact-zero effective weight.
+            # Inner FC1/FC2 loops then retain their unchecked hot path while
+            # caller-owned route tensors remain unchanged across graph replay.
+            staged_ids_ptr = cute.arch.alloc_smem(Int32, self.route_slots)
+            staged_weights_ptr = cute.arch.alloc_smem(Float32, self.route_slots)
+            staged_ids = cute.make_tensor(
+                staged_ids_ptr, cute.make_layout(self.route_slots)
+            )
+            staged_weights = cute.make_tensor(
+                staged_weights_ptr, cute.make_layout(self.route_slots)
+            )
+            route_slot = Int32(tidx)
+            while route_slot < Int32(self.route_slots):
+                raw_expert = Int32(topk_ids[route_slot])
+                expert = Int32(0)
+                weight = Float32(0.0)
+                if (
+                    raw_expert >= Int32(0)
+                    and raw_expert < Int32(cfg.weight_E)
+                ):
+                    expert = raw_expert
+                    weight = Float32(topk_weights[route_slot])
+                staged_ids[route_slot] = expert
+                staged_weights[route_slot] = weight
+                route_slot += Int32(self.launch_block_dim)
+            cute.arch.sync_threads()
+            topk_ids = staged_ids
+            topk_weights = staged_weights
         if cutlass.const_expr(self.compile_time_phase == 2):
             self._run_fc2(
                 Int32(bidx_x),
--- a/b12x/moe/_shared/kernels/micro.py
+++ b/b12x/moe/_shared/kernels/micro.py
@@ -554,6 +554,33 @@ class MoEMicroKernelBackend:
         return fp4_dot4_sum_f32acc(u_packed, x0, x1, x2, x3)
 
     @cute.jit
+    def _fc2_route(
+        self,
+        topk_ids: cute.Tensor,
+        topk_weights: cute.Tensor,
+        eid_addr: Int32,
+        route_expert_limit: Int32,
+    ) -> Tuple[Int32, Float32]:
+        """Return an addressable expert and its effective router weight.
+
+        Fixed-M fused launches receive a route table sanitized in shared
+        memory before FC1 and FC2 execute. The FC2-only endpoint instead uses
+        runtime M with a compile-time expert-capacity bucket, so it must check
+        each route against the resident expert count supplied by the caller.
+        Invalid routes address expert zero with an exact-zero effective weight.
+        """
+        raw_expert = Int32(topk_ids[eid_addr])
+        expert = raw_expert
+        weight = Float32(topk_weights[eid_addr])
+        if cutlass.const_expr(self.compile_time_phase == 2):
+            expert = Int32(0)
+            weight = Float32(0.0)
+            if raw_expert >= Int32(0) and raw_expert < route_expert_limit:
+                expert = raw_expert
+                weight = Float32(topk_weights[eid_addr])
+        return expert, weight
+
+    @cute.jit
     def _scale_byte_to_f32(self, byte: Uint32) -> Float32:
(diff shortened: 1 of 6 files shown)


voipmonitor:
@coderabbitai review

voipmonitor:
Published integration qualification used this PR head debaafe156c9824396178d53e01e5f15d2a2a04a in B12X tree ec6edd9da4687f83519fd37bd7322ea0800f0ace and image voipmonitor/vllm@sha256:ffb25774eaa90850b4cacfb88ed9e55072818e99bad977f1315c7118e7a730b2. The official Kimi-K3 MXFP4 plus Inferact DSpark TP16/DCP16 profile completed loading and graph capture with 1,033,126 physical target KV tokens. Its eight-run normalized median was 118.773 emitted tok/s and 31.430 target cycles/s; the 31.430 cycles/s acceptance-independent hot-path result establishes no runtime regression. No illegal-address or graph-replay error occurred. The maintainer merge map and complete evidence are in rtx6kpro#75.

lukealonso (changes requested):
Requesting changes because the static route staging also applies to `compile_time_phase=2`. `_compile_w4a16_fc2_direct()` configures `m=2`, `topk=1`, so the staged route capacity is two, but `_w4a16_fc2_direct_launch_flat()` passes the runtime `m` and `_run_fc2()` indexes through that runtime count. Existing FC2-only coverage uses three and seven routes, so routes beyond index one read past the staged shared tensors. Keep phase 2 off this static staging path or sanitize FC2 routes inline, then add invalid-ID eager and CUDA Graph coverage for `m=3` and `m=7`.

voipmonitor:
The FC2-only endpoint now validates each runtime route against the resident expert count. Invalid negative and upper-bound route IDs address expert zero with an exact-zero effective weight; caller-owned route tensors remain unchanged. Fixed-size shared route staging remains limited to fused launches whose route-table extent is known at compile time.

Validation on one RTX PRO 6000 Blackwell GPU with the Infernal Invocation CUDA 13.3 / Torch 2.13 environment:

- runtime route staging contract;
- native direct inactive-route handling for `M=1,2,4,8`;
- FC2-only invalid-route handling for narrow and wide `M=3` shapes;
- FC2-only eager and CUDA Graph replay at `M=7`;
- direct-kernel dispatch observation, finite/nonzero valid-route output, exact-zero inactive output, oracle parity, and immutable route inputs.

Result: `8 passed`; `ruff check` and `ruff format --check` pass for `tests/moe/test_w4a16_e2e.py`.

The implementation is in `af354efd`; `6a41770f` strengthens the dispatch and output evidence requested by automated review.

voipmonitor:
@lukealonso Your FC2 runtime-M finding is fixed in `af354efd`: phase-2 execution no longer uses the compile-time M=2 shared route table and validates each runtime route inline against the resident expert count. Direct-dispatch eager and CUDA Graph tests cover M=3/M=7, `-1`, `-9`, and the upper bound; Compute Sanitizer reports zero errors. The official Kimi-K3 MXFP4 + DSpark TP16/DCP16 profile loaded, captured, and measured 31.520 target cycles/s versus 31.430 for the comparison image. The PR body and tracked validation receipt now contain the exact commands, raw samples, and source identities. No standalone `.cu` source is introduced. Please re-review head `acd96cbf`.

lukealonso (changes requested):
Two blockers remain on `acd96cbf`:

1. **[P1] Update the generic compact-micro compiler call for the new route-limit argument.** `MoEMicroKernelBackend.__call__` now requires `route_expert_limit, m_val, grid_x, stream` after the barrier arguments, but `_get_micro_kernel()` in `b12x/moe/fused_moe/_impl.py` still sends the old `m_val, grid_x, stream` tail. Generic NVFP4/W4A8 micro compilation on a cache miss therefore shifts every tail argument and omits the required stream; the Trellis arguments are shifted too. Pass `Int32(weight_E)` before `compile_m`, update the cached compile signature or version if required, and cover compilation through the production generic-micro dispatch path.

2. **[P2] Validate `int64` route IDs before narrowing them to `Int32`.** In `b12x/moe/_shared/kernels/micro.py`, both `_fc2_route` and the fused shared-memory staging path perform `Int32(topk_ids[...])` before the bounds check. Public W4A16 paths accept `torch.int64`, so a value such as `2**32` truncates to expert 0 and contributes instead of being treated as inactive. Preserve the source width for validation, narrow only after a route is known valid, and add `int64` overflow coverage for fused and FC2-only execution.

The previous runtime-M FC2 staging issue is fixed.

voipmonitor:
The direct-micro compile ABI is qualified at `e38436d76a95c586c57e06646f8ea5b8c8ed11c7`.

`MoEMicroKernelBackend.__call__` accepts `route_expert_limit` before runtime M, grid extent, and CUDA stream. `_get_micro_kernel()` now supplies `weight_E` at that position. Without the argument, CuTe bound runtime M as the expert limit and bound the CUDA stream as `grid_x`; generic NVFP4 compilation then failed because the required `stream` argument was absent.

Regression coverage compiles the fused NVFP4 body and both split phases through the real CuTe compiler. The exact release composition combining this PR with B12X PR #223 produced:

```text
309 passed, 18 skipped in 123.46s
```

Covered suites:

```text
tests/gemm/test_trellis_k6_mcg_cute.py
tests/moe/test_w4a16_e2e.py
tests/moe/test_w4a16_mixed_trellis.py
tests/moe/test_w4a16_route_pack_warmup.py
tests/moe/test_moe_launch_param_regression.py
tests/test_packaging.py
```

The result covers native and Trellis routes, fused and split NVFP4 compilation, eager execution, CUDA Graph replay, mixed K3/K4/K5 dispatch, route packing, and packaged source availability. The durable validation record is `validation/performance/w4a16_inactive_routes_sm120.md`.

lukealonso (changes requested):
Follow-up on `e38436d`: the generic compact-micro compile ABI blocker is resolved by `97dfe7f`, which supplies `Int32(weight_E)` and adds fused and split-phase compile coverage.

One correctness blocker remains: `_fc2_route` and the fused shared-memory staging path in `b12x/moe/_shared/kernels/micro.py` still convert `topk_ids[...]` to `Int32` before checking its range. The public W4A16 entry points accept `torch.int64`, so an out-of-range value such as `2**32` aliases to expert 0 and contributes instead of being inactive. Validate at the source width, narrow only after the ID is known valid, and add `int64` overflow cases for fused and FC2-only execution.

voipmonitor:
The remaining source-width blocker is fixed at `98e3d95638776e78cf9b9de552345c121126b73b`.

Both the fixed-M shared-memory staging path and FC2-only runtime-M path now load route identifiers as `Int64`, validate against the resident expert range, and narrow to `Int32` only after validation. The non-FC2 hot path remains unchanged after its compile-time branch.

Validation on SM120 with CUDA 13.3, PyTorch 2.13, and CUTLASS DSL 4.6.2: nine targeted tests passed. Fused execution covers `int32` M=1/2/4/8 and `int64` M=1; FC2 covers `int32` and `int64` at M=3 for two intermediate widths. Both `int64` cases use `2**32`, and both eager execution and CUDA Graph replay require exact-zero inactive contributions and immutable route tensors.

voipmonitor:
Qualification for head `0eba6ae99e0d1fad6ec268d8c291f498ec1dd4d9`:

- the W4A16 routing and packaging suites passed 329 tests and skipped 18;
- coverage includes int32 and int64 route identifiers, fused and split execution, fixed and runtime M, eager execution, and CUDA Graph replay;
- official `moonshotai/Kimi-K3` MXFP4 with Inferact DSpark K7 ran on TP16/DCP16 from B12X tree `468c31256b585c8078782c63ba723e404c04eb76` in `voipmonitor/vllm@sha256:a54da4e2432138d42334cac54555f3a51188489cb66029da0d96e4b39162d726`;
- three streamed 512-token coding-decode samples measured 133.55, 118.72, and 91.05 tok/s (median 118.72 tok/s);
- no standalone native CUDA source is introduced.

The implementation applies to supported TP8, TP12, and TP16 W4A16 shapes. Inactive negative routes and out-of-range int64 route identifiers cannot alias a resident expert during narrowing.

lukealonso:
Please remove validation/performance/

voipmonitor:
Superseded by #247. `b12x/master` now has an in-kernel active-route predicate in the general microkernel path, so the fixed-M shared-memory staging added here would duplicate route validation. #247 is based on `a71c705f` and retains only the unresolved tensor-bound contract: FC2-only launches validate against the actual resident expert count rather than the compile capacity bucket, and all direct W4A16 route IDs are checked at source width before narrowing. Its focused eager and CUDA graph suite passed 11 cases on SM120. The replacement contains no `validation/performance` tree.

PR #228, opened 2026-08-17
[II] Ignore inactive MoE routes in tiny decode

voipmonitor:
## Resulting behavior

The B12X W4A8 tiny-decode MoE kernel treats every route whose expert ID is outside `[0, weight_E)` as inactive. Inactive routes use an in-range address for speculative reads and contribute zero to the intermediate and output tensors.

This matches the existing dynamic-MoE contract and vLLM scheduler semantics: CUDA-graph padding and non-local expert-parallel routes may carry expert ID `-1`. The kernel must not dereference that ID even when the corresponding router weight is expected to be zero.

## Technical reason

The tiny-decode FC1 and FC2 kernels previously formed weight addresses directly from `topk_ids`. A padded `-1` route therefore addressed memory before the expert allocation. Native KV offloading changes the allocation layout enough to make the invalid read reproducible during PIECEWISE CUDA-graph memory profiling, but neither the offload transport nor its filesystem tier is causal.

The implementation now:

- validates the route ID before forming expert weight addresses;
- substitutes expert zero only as a safe read address for inactive routes;
- suppresses FC1 intermediate writes and FC2 output accumulation for inactive routes;
- preserves the launch geometry and arithmetic of every valid route.

## Validation

- `ruff check` and `ruff format --check`: pass.
- `git diff --check`: pass.
- `tests/moe/test_cute_migration_moe_standard_corpus.py`: 5 passed on SM120.
- Added CUDA-graph replay coverage with a nonzero router weight on expert ID `-1`; output is compared with an oracle that explicitly masks inactive routes.
- DeepSeek-V4-Flash-0731, TP2/DCP1, DSpark K5, B12X A8, `MAX_NUM_BATCHED_TOKENS=8192`, native KV offload: all PIECEWISE memory profiles and model/DSpark/context-KV graph captures complete without `CUDA_LAUNCH_BLOCKING`.
- The same server completed an 85,986-token request and a strict required-tool request.
- Filesystem-tier replay read 1.60 GB across 795 block hits with no promotion/cascade failure or residual temporary file.

Configurations whose declared context exceeds the available KV budget now fail at the explicit capacity check rather than inside the tiny-decode kernel. That capacity failure is expected and is not bypassed by this change.

## Compatibility

Valid expert IDs retain the existing code path and launch policy. The only semantic change is for IDs that are outside the local expert range, which are defined as inactive.

### DIFF
--- a/b12x/moe/_shared/kernels/tiny_decode.py
+++ b/b12x/moe/_shared/kernels/tiny_decode.py
@@ -231,7 +231,14 @@ class MoETinyDecodeKernelBackend:
             rem = bidx % fc1_per_rt
             nt = rem // Int32(c["fc1_ktg"])
             ktg = rem % Int32(c["fc1_ktg"])
-            eid = Int64(Int32(topk_ids[rt_idx]))
+            route_eid = Int32(topk_ids[rt_idx])
+            route_active = route_eid >= Int32(0) and route_eid < Int32(c["weight_E"])
+            # vLLM represents CUDA-graph padding and non-local EP routes with
+            # an expert id outside [0, weight_E). Keep every address in range;
+            # inactive routes retain the pre-zeroed intermediate and output.
+            eid = Int64(0)
+            if route_active:
+                eid = Int64(route_eid)
             tok = rt_idx // Int32(c["num_topk"])
             we_base = w13_base + eid * Int64(c["w13_words"] * 4)
             se_base = sfb13_base + eid * Int64(c["sfb13_bytes"])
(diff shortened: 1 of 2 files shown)


lukealonso:
Review result: approved. The tiny-decode kernel bounds the expert ID before address formation and suppresses both FC1 intermediate writes and FC2 output accumulation for inactive routes. This preserves valid-route arithmetic and matches the pre-zeroed buffer contract under CUDA Graph replay. No blocking issues found.

PR #230, opened 2026-08-17
[II] Fix mapped W4A16 route namespace planning

voipmonitor:
## Behavior

W4A16 scratch planning sizes route metadata for both physical expert weights and the declared global router namespace. Packed projection tiers can therefore bind a contiguous `route_expert_map` whose length exceeds the number of locally stored expert weights.

## Technical reason

`route_expert_map` is indexed by global router IDs, while W4A16 kernels index physical weights after mapping. The route workspace must cover `max(weight_E, route_num_experts)`. Sizing packed plans only from `weight_E` rejects valid maps before execution; sizing only from `route_num_experts` can under-size plans whose declaration is narrower than their weight set.

## Compatibility

Plans without an expert map and plans with equal expert counts retain the same route dimensions. A wider declared namespace reserves the route metadata and launch geometry required by that map. For the GLM-5.2 shape (`H=6144`, `I=2048`, top-k 8), the measured scratch delta through 32 tokens is 0.001-0.021 MiB. At a 128-token capacity, route-block selection reduced scratch by 46.942 MiB for 32 local experts and 23.452 MiB for 64 local experts.

## Validation

- `tests/moe/test_fused_moe_planning.py`: 20 passed.
- `test_w4a16_scratch_plan_mapped_decode_is_graph_safe`: eager parity and CUDA graph replay passed on SM120.
- `ruff check b12x/moe/fused_moe/_impl.py tests/moe/test_fused_moe_planning.py`: passed.

### DIFF
--- a/b12x/moe/fused_moe/_impl.py
+++ b/b12x/moe/fused_moe/_impl.py
@@ -2591,7 +2591,9 @@ def _plan_core_workspace(
         )
         requested_route_E = int(route_num_experts or weight_E)
         full_rotation = weight_layout == "trellis_t256"
-        route_E = requested_route_E if full_rotation else int(weight_E)
+        # Route metadata must address every physical weight and every global
+        # router ID accepted through route_expert_map.
+        route_E = max(int(weight_E), requested_route_E)
         if full_rotation:
             if source_format not in _TRELLIS_SOURCE_FORMATS:
                 raise ValueError(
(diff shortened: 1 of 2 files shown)


lukealonso (approved):
The planner now covers both the physical weight namespace and the declared global router namespace while retaining `weight_E` for physical storage. The narrower-router and wider-router cases establish the required `max(weight_E, route_num_experts)` contract. Approved.

PR #231, opened 2026-08-18
feat(attention): add KVarN packed-latent staging and native reader

JMPSequeira:
New attention.kvarn_mla op: Triton staging of KVarN K2/K4/K5 G64 tiles (packed latent + exact BF16 RoPE) into the canonical FP8 record format, compact per-rank DCP wire staging/materialization, and native packed decode without a global row stage, backed by KVarN decode grids, math, SMEM layouts, and traits in _shared/mla. Registered in the op registry with reference tests for the staging and round-trip contracts.

## Summary

Adds `attention.kvarn_mla` for packed-latent KVarN attention on SM120/SM121.

- Stages K2/K4/K5 and exact BF16 tiles into canonical FP8 records with exact BF16 RoPE support.
- Compacts and materializes per-rank DCP history with Int64 offset calculations.
- Adds native packed K5 decoding with exact-pool and Hadamard modes.
- Adds KVarN decode math, shared-memory layouts, traits, grids, validation, and launch configuration.
- Handles exact-page sink and tail windows without invalid exact-page mappings.
- Registers the operation and exposes capability detection and staging APIs.
- Preserves existing decode paths through shared generalized staging and chunk logic.
- Adds contract tests for staging, compact wire geometry, round-trip materialization, SM120 decoding, large offsets, and exact-page eligibility.

### DIFF
--- a/b12x/__init__.py
+++ b/b12x/__init__.py
@@ -43,6 +43,7 @@ _OPS: tuple[str, ...] = (
     "attention.dense_mla",
     "attention.sparse_mla",
     "attention.compressed_mla",
+    "attention.kvarn_mla",
     "attention.nsa_indexer",
     "attention.varlen",
     "comm.pcie",
--- a/b12x/attention/__init__.py
+++ b/b12x/attention/__init__.py
@@ -6,8 +6,9 @@
   optional causal sliding-window masking.
 - ``sparse_mla``: top-k-selected MLA, including strided physical records.
 - ``compressed_mla``: MLA decode directly from compressed KV pages (DSV4).
+- ``kvarn_mla``: KVarN packed-latent (K2/K4/K5 G64) staging and native
+  sparse-MLA decode without a global row stage.
 - ``nsa_indexer``: the NSA index stage — quantize -> score -> select.
-- ``varlen``: contiguous batched/varlen attention (reduced-assurance tier).
 """
 
 from __future__ import annotations
@@ -20,6 +21,7 @@ _OP_MODULES = (
     "dense_mla",
     "sparse_mla",
     "compressed_mla",
+    "kvarn_mla",
     "nsa_indexer",
     "varlen",
 )
--- a/b12x/attention/_shared/mla/decode_math.py
+++ b/b12x/attention/_shared/mla/decode_math.py
@@ -989,6 +989,27 @@ def s1_qk_nope_nvfp4_bf16(
 
 
 @cute.jit
+def s1_qk_nope_kvarn_bf16(
+    qk, q_base: Int32, kv_base: Int32, warp_first_cand: Int32, lane: Int32,
+    *, num_scales: cutlass.Constexpr, quant_tile: cutlass.Constexpr,
+    q_nope_bf16_stride: cutlass.Constexpr, kv_smem_stride: cutlass.Constexpr,
+):
+    gid = lane >> Int32(2)
+    tid = lane & Int32(3)
+    a_row = (lane & Int32(7)) + ((lane >> Int32(3)) & Int32(1)) * Int32(8)
+    a_col = (lane >> Int32(4)) * Int32(8)
+    row = kv_base + (warp_first_cand + gid) * Int32(kv_smem_stride)
+    for blk in cutlass.range_constexpr(num_scales):
+        for ks in cutlass.range_constexpr(quant_tile // 16):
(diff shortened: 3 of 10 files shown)


JMPSequeira:
Review response — all five findings addressed in a09cc5d (7/7 kvarn tests pass on RTX PRO 6000 Blackwell / SM120, torch 2.12.1+cu130, triton 3.7.1, nvidia-cutlass-dsl 4.6.0; tests/test_registry.py, test_packaging.py, tests/lint/ also pass — the remaining registry failure is the pre-existing `quantization.mxfp6` mismatch on master, untouched here).

**1. CRITICAL decode_math.py:810 — valid, fixed.** The port had dropped both KVarN math functions as "unused" (the usage search missed that the callers `s1_qk_nope_block_scaled`/`s6_xv_nope` live in the same module), leaving the orphaned `@cute.jit` decorators you flagged. Both are restored verbatim from the deployment source, so `decode_math` is again an exact port. Why the original tests passed: they cover only the Triton staging surface; the KVARN_K5/KVARN_EXACT_BF16 decode specializations were never traced on the prep host (no CUTLASS DSL wheel). Closed both gaps: installed nvidia-cutlass-dsl 4.6.0 and added `test_native_packed_decode_matches_reference`, which runs `native_packed_k5_decode` through the real CuTeDSL SM120 grid against an fp32 reference. Verified it fails with exactly `NameError: name 's1_qk_nope_kvarn_bf16' is not defined` on the pre-fix tree and passes after (cos ≥ 0.9995, LSE within 0.5 of the fp32 logsumexp).

**2. CRITICAL api.py:170 — valid, fixed.** Every identifier-to-byte-offset product in `_stage_k5_as_fp8_records_kernel`, `_stage_compact_kvarn_native_packed_kernel`, `_stage_compact_kvarn_native_exact_kernel`, and `_materialize_compact_kvarn_native_records_kernel` now widens the identifier (`block`, `page`, `pool_slot`, `safe_pool_slot`, `exact_id`, `rank`, `row`, `output_row`) to Int64 before multiplying by the tile/record strides, per the repo's 64-bit pool-addressing rule. Added `test_scaled_offsets_survive_int32_boundary` with live tail indices past `INT32_MAX / stride`: a 70,000-block K5 arena (block 69,999 × 30,848 B), a 100,000-page compact wire (page 99,999 × 26,752 B), and a two-rank gather whose rank-1 wire base (1 × 2,675,255,360 B) exceeds 2³¹ — for both packed and exact-pool paths, checked against the reference records. The test fails on the Int32 build and passes after the widening.

**3. MAJOR api.py:312 — valid, fixed.** `_native_exact_id` now returns `(in_window, exact_id)` where `in_window = mapped & 0 <= within_req < 16`; both staging kernels fold it into `exact` before the index/latent/RoPE stores, so pool-marked body pages stage as packed instead of writing a negative or aliased exact id when `tl.device_assert` is compiled out. Added `test_exact_pages_outside_sink_tail_window_stay_packed` (page_len=20, pool-marked local pages 2 and 5): on the pre-fix tree it reproduces your analysis exactly (`index[2] == -2`); after the fix both stay `-1`, keep their packed tiles verbatim, and only the true sink page writes the exact region.

**4. MINOR kernel.py:137 — applied** your present-state rewrite verbatim.

**5. MINOR api.py:756 — fixed** (`stride 26752`), and the adjacent `v18` cache-version comment in kernel.py now reads `v19` to match the literal.

One deliberate contract change in the tests: staged-record comparison now asserts dequantized latent equality within one E4M3 step instead of bit-exact codes — the kernel's `amax / 448` fp32 division can round one ULP away from torch, which legitimately flips a single code at a rounding tie. RoPE stays bit-exact and scales stay within 2e-7 relative.

JMPSequeira:
Addressed in 17d0ef6: `test_scaled_offsets_survive_int32_boundary` now skips when `torch.cuda.mem_get_info` reports under 10 GiB free device memory (verified peak: 2.16G K5 arena + 2.68G wire + 5.35G two-rank gather ≈ 10.2 GB plus context) or, when psutil is importable, under 4 GiB available host RAM for the 2.68G wire reference copy. The psutil check degrades to a no-op rather than adding a dependency. Boundary coverage unchanged; full kvarn battery 7/7 on RTX PRO 6000 Blackwell (guard passes through, 96 GiB free).

lukealonso:
@JMPSequeira Do you have this working E2E in vLLM? Do you have some performance/quality metrics?

PR #247, opened 2026-08-27
fix(moe): bound native W4A16 routes by resident tensors

voipmonitor:
## Behavior

Native ModelOpt W4A16 microkernels treat an expert identifier outside the resident expert tensor extent as an inactive route. An inactive route addresses no absent expert data and contributes exactly zero. Route identifiers are validated at their source width before conversion to the 32-bit kernel address type.

The FC2-only endpoint compiles against a reusable expert-capacity bucket but validates each launch against the actual resident expert count. Fused FC1+FC2 launches retain the route predicate already used by `b12x/master` and validate against their exact compile-time expert count. Caller-owned route identifiers and weights remain unchanged in eager execution and CUDA graph replay.

## Technical reason

The FC2-only compile bucket can exceed the number of experts represented by the supplied weight and scale tensors. Using the bucket as the route bound permits an identifier to address beyond those tensors. Validating an `int64` identifier after narrowing also permits values such as `2**32` to alias expert zero.

The launch ABI now carries the actual expert count. FC2 kernels resolve every route against that count, while fused kernels use their exact configured extent. Both paths convert an identifier to `Int32` only after the range check succeeds.

## Compatibility

- Scope: native direct W4A16 microkernel execution.
- Launch geometry, workspace allocation, numerical reduction order, and valid-route arithmetic are unchanged.
- Dynamic route packing and the tiny-decode kernel retain their existing contracts.
- Inactive arithmetic slots remain in fixed CUDA graph shapes and contribute zero.
- No runtime allocation or host synchronization is added.

## Validation

Status: **qualified** for the focused native W4A16 route contract on one NVIDIA RTX PRO 6000 Blackwell GPU with CUDA 13.3 and PyTorch 2.13.

- `ruff check` passed for all changed Python files.
- Focused Pytest result: `11 passed in 6.54s`.
- Fused tests cover rows 1, 2, 4, and 8 with `int32` route tensors and row 1 with `int64` route tensors.
- FC2-only tests cover runtime row count 3, `int32` and `int64` identifiers, and intermediate widths 256 and 512.
- Assertions cover eager execution, CUDA graph capture and replay, exact-zero inactive output, finite valid output, reference parity, and immutable route tensors.
- `int64` coverage includes `2**32`, which must remain inactive rather than aliasing expert zero.

No service-level performance claim is made by this PR. A composed runtime will be measured separately without changing this source contract.

## AI assistance

AI assistance was used to implement and test the change. The submitting human must review every changed line and be able to defend the behavior before merge.

## Summary

Native direct W4A16 MoE routes now validate expert identifiers against resident expert tensor extents. Invalid identifiers become inactive routes and produce exact-zero output. Validation occurs at the source route width before conversion to `Int32`.

FC2-only execution uses the runtime resident expert count. Fused FC1+FC2 execution retains its configured expert extent. Route identifiers and weights remain unchanged during eager execution and CUDA graph replay.

The change preserves launch geometry, workspace allocation, reduction order, valid-route arithmetic, dynamic route packing, tiny-decode contracts, and the direct microkernel compile ABI. It adds no runtime allocation or host synchronization.

## Validation

- Ruff checks passed.
- 11 focused Pytest cases passed.
- Coverage includes eager execution, CUDA graph replay, route-width validation, invalid identifiers including `2**32`, exact-zero inactive output, reference parity, and immutable route inputs.
- Validation ran on an NVIDIA RTX PRO 6000 Blackwell GPU with CUDA 13.3 and PyTorch 2.13.


voipmonitor:
@coderabbitai review

PR #249, opened 2026-08-28
feat(gdn): add direct single-token KDA decode

MadeBy561:
## Purpose

Add a native, graph-safe B12X KDA specialization for ordinary decode where every live request contributes exactly one token. It consumes caller-owned tensors directly and compiles out packed-metadata validation and staging work.

The API is explicit about its trusted scheduler contract. Generic packed, speculative, padded, and multi-token requests continue through `run_kda` with the existing device-side validation. No FLA or Torch fallback is added. No existing open B12X PR covers this KDA path.

## Test plan and result

- `pytest -q tests/sequence/test_gdn_decode_kda.py`
- Result: `16 passed` on RTX PRO 6000 Blackwell. Coverage includes BF16/FP32 recurrent states, exact reference comparison, strided beta, null state, duplicate-slot transactionality, CUDA-graph replay, `torch.compile`, and state offsets beyond the signed-32-bit element boundary.
- Matched GLM-5.3 NVFP4 TP4 MTP0 normal-sampling C1, when added to the separately measured M=1 MoE specialization: `128.3 -> 140.58 tok/s` at context 0 and `127.8 -> 139.97 tok/s` at 8k. Stock JJ was `123.2/122.7 tok/s`; qualified r19 was `141.0-141.1/~140.6 tok/s`.

This PR changes neither KDA math nor precision. The final serving stack remained native B12X sparse MLA, KDA, A4 MoE, and PCIe all-reduce.

Assisted by OpenAI Codex; the submitter reviewed the measured behavior and final change.

## Summary

Adds `run_kda_single_token` as a native, graph-safe B12X KDA fast path for ordinary single-token decode.

- Uses caller-owned tensors directly.
- Skips packed-metadata validation and staging for the trusted single-token path.
- Preserves existing `run_kda` behavior for packed, speculative, padded, and multi-token requests.
- Keeps KDA math and precision unchanged.
- Adds validation for shapes, devices, dtypes, capacity, output aliasing, and numeric parameters.
- Adds CUDA graph replay and strided-beta coverage.

All 16 tests in `tests/sequence/test_gdn_decode_kda.py` pass, including reference accuracy, recurrent-state variants, duplicate-slot transactionality, `torch.compile`, CUDA graph replay, and large state offsets.

Measured GLM-5.3 NVFP4 TP4 MTP0 normal-sampling C1 throughput increased from 128.3/127.8 to 140.58/139.97 tok/s at context 0/8k when combined with a separately measured M=1 MoE specialization.

### DIFF
--- a/b12x/sequence/gdn_decode/__init__.py
+++ b/b12x/sequence/gdn_decode/__init__.py
@@ -57,6 +57,7 @@ META = OpMeta(
         "bind_kda",
         "run",
         "run_kda",
+        "run_kda_single_token",
         "reference",
         "is_supported",
     ),
@@ -96,6 +97,7 @@ if TYPE_CHECKING:
         reference,
         run,
         run_kda,
+        run_kda_single_token,
     )
 
 install_lazy_api(globals(), META)
--- a/b12x/sequence/gdn_decode/_impl.py
+++ b/b12x/sequence/gdn_decode/_impl.py
@@ -851,6 +851,126 @@ def run_kda(
     return binding.output
 
 
+def run_kda_single_token(
+    binding: KdaBinding,
+    *,
+    mixed_qkv: torch.Tensor,
+    raw_g: torch.Tensor,
+    raw_beta: torch.Tensor,
+    z: torch.Tensor,
+    state_indices: torch.Tensor,
+    output: torch.Tensor,
+    lower_bound: float = -5.0,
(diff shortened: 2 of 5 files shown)


lukealonso:
I would rather not add another public API entrypoint for this. I'm amenable to a fully internal (routed) kernel optimization though.

PR #253, opened 2026-08-28
perf(gdn): support direct KDA binding

MadeBy561:
## Summary

- Support direct KDA binding through the existing `bind_kda` / `run_kda` API when caller-owned token and request buffers are smaller than the plan.
- Keep planned capacities as kernel maxima while reduced capacities drive launch grids and masks through runtime, `do_not_specialize` arguments.
- Preserve packed-layout validation, transactional failure, fixed state-column geometry, and CUDA graph replay.

This adds no public entry point or serving flag. "Reduced-capacity binding" describes the shape contract; the optimization is direct binding of caller buffers, not a single-token-only path.

## Contract

For reduced request-capacity bindings, every bound request row is live, `num_seqs` must equal that capacity, and `query_start_loc[bound_seqs]` must equal the independent device-side `num_tokens` scalar. Full-capacity bindings retain the existing behavior where both live counts may change between graph replays.

`raw_beta` may be row-strided because kernels already consume its explicit strides. Token and request capacities must fit the plan, state columns must exactly match the plan, and token capacity must fit the request-column envelope.

Only the reduced-capacity policy boolean specializes. Planned capacities and static model geometry remain compile-time; live request, token, and accepted-prefix counts do not enter compile/cache keys.

## Why

Speculative verification already owns exact-size projected tensors and accepted-prefix metadata. Binding those buffers to an existing plan lets integrations avoid plan-sized per-layer staging copies without moving planner policy into the caller. The vLLM consumer is local-inference-lab/vllm#504.

## Correctness and graph tests

Run on SM120 in the serving image:

```bash
/opt/venv/bin/python -m pytest tests/sequence/test_gdn_decode_kda.py -q
```

Result: `16/16 passed`.

Coverage includes:

- reference agreement for reduced-capacity output and recurrent state;
- two distinct reduced request/token capacities under frozen kernel resolution, with unchanged Triton cache cardinality;
- strided `raw_beta` on the reduced-capacity route;
- reduced-capacity CUDA graph capture/replay;
- transactional rejection of both a mismatched request count and `query_start_loc[bound_seqs] != num_tokens`, with all-NaN output and unchanged recurrent state;
- full-capacity replay with changing live counts; and
- the existing rejected-draft, duplicate-slot, null-state, large-offset, and `torch.compile` coverage.

The linked vLLM test passes `1 passed, 38 deselected` and verifies that its direct route supplies `num_tokens` from a dedicated live-count buffer rather than aliasing the terminal query offset.

Static checks:

```bash
uvx ruff check --ignore F841 b12x/sequence/gdn_decode/_impl.py b12x/sequence/gdn_decode/_kernels.py tests/sequence/test_gdn_decode_kda.py
git diff --check 2fcf23a0ce269be27b2e03fece73d46e90e6aeea...3c485daaa0140bf00d0172c55e3af83e445e2de5
```

Both passed. `F841` is ignored only for an unrelated baseline unused variable already present in the test file.

## Performance evidence

An earlier matched staged/direct-binding A/B isolated the staging-removal mechanism before this internal API refactor. Both arms used the same GLM-5.3-Flash-NVFP4 TP4 stack, W4A4 target MoE, probabilistic MTP3, FP8 KV, C32 capacity, 1M model length, 8,192-token batching, CUDA graphs capped at 128 rows, and normal model sampling.

| Arm | C1 delivered tok/s | C1 verifier steps/s | C30 aggregate tok/s | C30 verifier steps/s |
|---|---:|---:|---:|---:|
| staged control | 174.5 | 70.59 | 1,510.2 | 608.07 |
| direct binding, run 1 | 167.7 | 74.35 | 1,527.5 | 618.09 |
| direct binding, run 2 | 184.6 | 74.00 | 1,526.0 | 609.52 |

The isolated A/B supports the mechanism: C1 verifier work increased about 5.1%, while delivered throughput varied with stochastic acceptance; mean C30 delivered throughput increased about 1.1%.

The exact final heads were requalified after the terminal-offset fix. This is integration qualification, not isolated attribution to this PR:

| Run | Delivered tok/s | Verifier steps/s | Mean accepted length |
|---|---:|---:|---:|
| C1, final heads | 199.3 | 79.82 | 2.50 |

No temperature override was used. Ratio direction is candidate/control, higher is better.

Raw-result SHA-256:

- isolated staged control: `fde1050b574af48e31dece2ed3908730e78de33342e2df2ec1f6301d9ed780a4`
- isolated direct run 1: `42f4690dc6bbca86f194de02f748e5907d74c7b34458f7ead4e874fcc5dcd7b2`
- isolated direct run 2: `1119f71dc5a4cdbdd2d9fe198ce3b966d1dce1198ddcba86b8dc35f373b7ed72`
- final-head C1: `abc8891fcec2b1a98cbb90d0d5ea3520248fe8a31a40a1a3a26404558f3e1148`

Benchmark client: `llm_decode_bench` v0.4.29 at `29b8f0ecd16b061114adc3610a95e8f7a1979a69`.

## Provenance

- B12X base: `2fcf23a0ce269be27b2e03fece73d46e90e6aeea`
- B12X head: `3c485daaa0140bf00d0172c55e3af83e445e2de5`
- vLLM #504 qualification head: `71ad9871ed6574cd559ffbe123843c48013ea7c9`
- base serving image ID: `sha256:aef068522bbcfd6f0e4b865d39f8aca80d090af813c65960d743a0486d87f48c`; the exact PR source files were bind-mounted over that image.

The qualification box used four NVIDIA RTX PRO 6000 Blackwell Max-Q Workstation Edition GPUs at 300 W. During C1 all four were P1, memory clock was 16,365 MHz, graphics clocks were 2,265-2,302 MHz, and the active throttle mask was zero.

## Duplicate-work check

- Closed #249 used additional public single-token entry points; this PR supersedes that design with internal direct binding through the existing API.
- vLLM #490 integrates the old API and does not cover speculative verified prefixes.
- No other open B12X PR found by the current title/body search implements reduced-capacity direct KDA binding.
- #252 is unrelated and is not a dependency.

OpenAI Codex assisted with implementation, tests, benchmarking, and PR preparation. The human submitter must review and be able to defend every changed line before merge.

## Summary

Adds graph-safe reduced-capacity KDA binding through the existing `bind_kda` / `run_kda` contract. Caller-owned token and request tensors can be smaller than planned capacity without staging. Planned capacities remain compile geometry. Runtime counts control launch grids and masks.

## Technical changes

- Preserves planned state-index dimensions while deriving runtime bounds from bound tensors.
- Adds bound-aware validation and launch sizing.
- Retains packed-metadata, duplicate-slot, range, padding, recurrent-state, and transactional validation.
- Supports non-contiguous `raw_beta`.
- Preserves CUDA graph capture and replay with changing live counts.
- Supports verified-prefix speculative transactions used by MTP/DFlash.
- Adds exact-capacity kernel handling to avoid unnecessary metadata loads.
- Adds no public API entrypoint or flag.

## Compatibility

Existing `bind_kda` / `run_kda` behavior remains available for packed, padded, speculative, and multi-token requests. Planned compile and cache geometry remains stable.

## Validation

- 16 KDA tests pass.
- Coverage includes reduced-capacity reference agreement, kernel-cache stability, CUDA graph replay, changing live counts, and transactional rollback.
- Lint and diff checks pass.
- Serving qualification covers the revised B12X and vLLM stack with C1 and C30 workloads.

### DIFF
--- a/b12x/sequence/gdn_decode/_kernels.py
+++ b/b12x/sequence/gdn_decode/_kernels.py
@@ -23,35 +23,56 @@ def _reset_validation_kernel(
         tl.store(error_code, 0)
 
 
-@triton.jit
+@triton.jit(do_not_specialize=["bound_seqs", "bound_tokens"])
 def _validate_packed_metadata_kernel(
     query_start_loc,
     num_accepted_tokens,
     num_seqs,
     num_tokens,
+    bound_seqs,
+    bound_tokens,
     error_code,
     MAX_SEQS: tl.constexpr,
     MAX_TOKENS: tl.constexpr,
     STATE_INDEX_COLUMNS: tl.constexpr,
+    EXACT_SEQS: tl.constexpr,
 ):
     request = tl.program_id(0)
-    live_seqs = tl.load(num_seqs).to(tl.int32)
+    declared_seqs = tl.load(num_seqs).to(tl.int32)
+    if EXACT_SEQS:
+        live_seqs = bound_seqs
+    else:
+        live_seqs = declared_seqs
     live_tokens = tl.load(num_tokens).to(tl.int32)
 
     if request == 0:
         counts_invalid = (
-            (live_seqs < 0)
-            | (live_seqs > MAX_SEQS)
+            (bound_seqs < 0)
+            | (bound_seqs > MAX_SEQS)
+            | (bound_tokens < 0)
+            | (bound_tokens > MAX_TOKENS)
+            | (live_seqs < 0)
+            | (live_seqs > bound_seqs)
             | (live_tokens < 0)
-            | (live_tokens > MAX_TOKENS)
+            | (live_tokens > bound_tokens)
         )
         first = tl.load(query_start_loc).to(tl.int32)
-        safe_last = tl.maximum(0, tl.minimum(live_seqs, MAX_SEQS))
-        last = tl.load(query_start_loc + safe_last).to(tl.int32)
-        if counts_invalid | (first != 0) | (last != live_tokens):
-            tl.atomic_or(error_code, 2)
+        if EXACT_SEQS:
+            last = tl.load(query_start_loc + bound_seqs).to(tl.int32)
+            if (
+                counts_invalid
+                | (declared_seqs != bound_seqs)
+                | (first != 0)
+                | (last != live_tokens)
+            ):
+                tl.atomic_or(error_code, 2)
+        else:
+            safe_last = tl.maximum(0, tl.minimum(live_seqs, bound_seqs))
+            last = tl.load(query_start_loc + safe_last).to(tl.int32)
+            if counts_invalid | (first != 0) | (last != live_tokens):
+                tl.atomic_or(error_code, 2)
 
-    if request < tl.maximum(0, tl.minimum(live_seqs, MAX_SEQS)):
+    if request < tl.maximum(0, tl.minimum(live_seqs, bound_seqs)):
         start = tl.load(query_start_loc + request).to(tl.int32)
         end = tl.load(query_start_loc + request + 1).to(tl.int32)
         accepted = tl.load(num_accepted_tokens + request).to(tl.int32)
@@ -68,12 +89,13 @@ def _validate_packed_metadata_kernel(
             tl.atomic_or(error_code, 2)
 
 
-@triton.jit
+@triton.jit(do_not_specialize=["bound_seqs"])
 def _validate_active_state_slots_kernel(
     query_start_loc,
     num_accepted_tokens,
     state_indices,
     num_seqs,
+    bound_seqs,
     duplicate_slots,
     error_code,
     stride_indices_request: tl.constexpr,
@@ -84,12 +106,16 @@ def _validate_active_state_slots_kernel(
     TABLE_SIZE: tl.constexpr,
     HAS_NULL_STATE_INDEX: tl.constexpr,
     NULL_STATE_INDEX: tl.constexpr,
+    EXACT_SEQS: tl.constexpr,
 ):
     cell = tl.program_id(0)
     request = cell // STATE_INDEX_COLUMNS
     column = cell % STATE_INDEX_COLUMNS
-    live_seqs = tl.load(num_seqs).to(tl.int32)
-    if request >= tl.maximum(0, tl.minimum(live_seqs, MAX_SEQS)):
+    if EXACT_SEQS:
+        live_seqs = bound_seqs
+    else:
+        live_seqs = tl.load(num_seqs).to(tl.int32)
+    if request >= tl.maximum(0, tl.minimum(live_seqs, bound_seqs)):
         return
 
     start = tl.load(query_start_loc + request).to(tl.int32)
@@ -140,7 +166,7 @@ def _validate_active_state_slots_kernel(
         tl.atomic_or(error_code, 1)
 
 
-@triton.jit
+@triton.jit(do_not_specialize=["bound_seqs"])
 def _packed_sequential_kda_decode_kernel(
     mixed_qkv,
     a,
@@ -152,6 +178,7 @@ def _packed_sequential_kda_decode_kernel(
     num_accepted_tokens,
     state_indices,
     num_seqs,
+    bound_seqs,
     output,
     error_code,
     scale,
@@ -179,6 +206,7 @@ def _packed_sequential_kda_decode_kernel(
     QK_L2NORM: tl.constexpr,
     HAS_NULL_STATE_INDEX: tl.constexpr,
     NULL_STATE_INDEX: tl.constexpr,
+    EXACT_SEQS: tl.constexpr,
 ):
     value_tile = tl.program_id(0)
     request_value_head = tl.program_id(1)
@@ -186,8 +214,11 @@ def _packed_sequential_kda_decode_kernel(
     value_head = request_value_head % VALUE_HEADS
     key_head = value_head
 
-    live_seqs = tl.load(num_seqs).to(tl.int32)
-    if (request >= tl.maximum(0, tl.minimum(live_seqs, MAX_SEQS))) | (
+    if EXACT_SEQS:
+        live_seqs = bound_seqs
+    else:
+        live_seqs = tl.load(num_seqs).to(tl.int32)
+    if (request >= tl.maximum(0, tl.minimum(live_seqs, bound_seqs))) | (
         tl.load(error_code).to(tl.int32) != 0
     ):
         return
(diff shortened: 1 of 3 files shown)


MadeBy561:
@lukealonso Reworked this around the existing `bind_kda` / `run_kda` contract as requested on #249. There is no new public entrypoint or flag. Reduced token/request capacities are routed internally; live counts are runtime `do_not_specialize` arguments, while the plan remains the compile geometry.

The revised KDA suite passes 16/16, including reference agreement, transactional mismatch handling, reduced-capacity graph replay, and changing live counts under a full-capacity captured graph. The exact revised stack also requalified at 201.9 tok/s C1 on one normal-sampling run and 1,579.2 tok/s aggregate at C30; a second C1 run had lower stochastic acceptance but the same ~79.8 verifier steps/s.

b12x/sequence/gdn_decode/_kernels.py
```diff
@@ -23,35 +23,50 @@ def _reset_validation_kernel(
...
+            | (live_tokens > bound_tokens)
         )
         first = tl.load(query_start_loc).to(tl.int32)
-        safe_last = tl.maximum(0, tl.minimum(live_seqs, MAX_SEQS))
-        last = tl.load(query_start_loc + safe_last).to(tl.int32)
-        if counts_invalid | (first != 0) | (last != live_tokens):
-            tl.atomic_or(error_code, 2)
-
-    if request < tl.maximum(0, tl.minimum(live_seqs, MAX_SEQS)):
+        if EXACT_SEQS:
```
lukealonso, line 60: Reduced request-capacity bindings still need the terminal packed-layout invariant. In this branch, `query_start_loc[bound_seqs] < num_tokens` passes all per-request checks, leaving trailing live tokens unowned; recurrence writes only through the terminal offset while downstream work is driven by `num_tokens`. Please reject `query_start_loc[bound_seqs] != num_tokens` before any state mutation.

The linked vLLM direct-binding path currently passes `num_tokens=query_start_loc[num_requests:num_requests+1]`, so that comparison would be tautological there. It should supply an independent live-token scalar, for example its existing `_b12x_kda_num_tokens` buffer, so B12X can enforce this invariant end to end.

tests/sequence/test_gdn_decode_kda.py
```diff
@@ -373,6 +374,104 @@ def test_glm53_tp8_head_geometry_matches_reference() -> None:
     )
 
 
+def test_kda_subcapacity_matches_reference_and_replays_cuda_graph() -> None:
```
lukealonso, line 385: This proves replay for one reduced capacity, but not the cache/capture invariant introduced by `do_not_specialize`. Please exercise at least two distinct reduced request/token capacities under frozen kernel resolution, or add an equivalent cache assertion, so a bound capacity accidentally entering Triton specialization is caught. It would also be useful to make `raw_beta` strided in one reduced-capacity case, since `bind_kda` now explicitly permits that layout.

lukealonso:
Overall, this is the right direction. The change is best understood as **direct KDA binding**: the existing `bind_kda` / `run_kda` path accepts caller-owned, reduced-capacity tensors, retains planned capacities as the kernel maxima, and internally shrinks the launch grids. The vLLM consumer can then skip plan-sized staging copies without adding another public API. That is a better description than “single-token,” since the contract supports multiple requests and tokens and the optimization is fundamentally about binding caller buffers directly.

The missing terminal-offset validation in the inline comment blocks merge for me, particularly because the current linked vLLM path aliases `num_tokens` to the terminal query offset. Once B12X restores that invariant, vLLM supplies an independent live-token scalar, and the regression case is covered, the overall design looks sound.

I suggest renaming the PR to `perf(gdn): support direct KDA binding`, while using “reduced-capacity binding” to describe the shape contract. The earlier isolated A/B supports the mechanism, but the final-head full-stack numbers are integration qualification rather than isolated attribution. The generated tail of the PR description should also be corrected: it says 20/20 tests while the authored section and current parametrization indicate 16.

b12x/sequence/gdn_decode/_kernels.py
```diff
@@ -23,35 +23,50 @@ def _reset_validation_kernel(
...
+            if counts_invalid | (declared_seqs != bound_seqs) | (first != 0):
+                tl.atomic_or(error_code, 2)
+        else:
+            safe_last = tl.maximum(0, tl.minimum(live_seqs, bound_seqs))
+            last = tl.load(query_start_loc + safe_last).to(tl.int32)
+            if counts_invalid | (first != 0) | (last != live_tokens):
+                tl.atomic_or(error_code, 2)
```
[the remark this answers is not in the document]
MadeBy561: Fixed in B12X head `3c485da`. Reduced request-capacity validation now loads `query_start_loc[bound_seqs]` and rejects it when it differs from the independent `num_tokens` scalar before recurrence. The regression exercises `[0, 2, 3]` with `num_tokens=4` and verifies metadata error signaling, all-NaN output, and unchanged recurrent state.

b12x/sequence/gdn_decode/_kernels.py
MadeBy561: Fixed end to end in B12X `3c485da` and vLLM #504 head `71ad9871e`. B12X restores the terminal packed-layout invariant in reduced-capacity mode. vLLM now fills its dedicated `_b12x_kda_num_tokens` buffer from `mixed_qkv.shape[0]` and passes that buffer, so the check no longer aliases the terminal query offset. The focused vLLM test asserts both the value and non-aliasing.

tests/sequence/test_gdn_decode_kda.py
MadeBy561: Added in `3c485da`. The test binds 2-request/5-token and 3-request/6-token capacities to the same 4-request/8-token plan, freezes kernel resolution for the second launch, and asserts all five Triton kernel cache cardinalities remain unchanged. Both cases use the same strided `raw_beta` layout. The exact final KDA file passes 16/16 on SM120.

MadeBy561:
Updated per review in `3c485da`: renamed/reframed as direct KDA binding, restored the terminal packed-layout invariant, linked the independent vLLM live-token scalar in #504, and added the requested two-capacity frozen-cache plus strided-`raw_beta` coverage. The exact final SM120 KDA file passes 16/16; the final integrated C1 qualification is 199.3 tok/s at 79.82 verifier steps/s. Please let me know if you want the contract expressed differently.

MadeBy561:
Marking draft after a new multi-request DFlash2 correctness gate exposed a real integration regression in the #253 + vLLM #504 direct-binding stack.

Matched K8, probabilistic/standard-rejection GSM8K controls on the same TP4 server:

- direct binding at C4: acceptance length 1.05-2.05; some requests repeated punctuation to the output cap
- staged control at C4: acceptance length 5.79; 68.4% draft acceptance; 4/4 requests terminated normally
- staged control at C32: acceptance length 5.53; 64.7% draft acceptance; all 32 requests remained coherent

Removing the new KV page/group candidates did not fix it; removing the #253/#504 pair did. I am isolating reduced-capacity kernel execution from the vLLM CUDA-graph/live-prefix contract now. The PR will stay draft and the performance claims will be requalified only after a batched per-request-prefix regression test and the serving gate pass.

PR #256, opened 2026-08-29
fix(moe): keep route workspace bounds runtime-dynamic

malaiwah:
## Problem

Triton specializes unannotated integer arguments. Route-packing requests can therefore compile new `_w4a16_route_count_kernel`, `_pack_topk_routes_post_prefix_kernel`, and `_pack_topk_routes_sort_kernel` variants when the live route count or preallocated workspace capacity changes. In a server that treats inference-time compilation as a warning, this produces a warning on the first unmatched request shape.

## Change

- Keep post-prefix workspace bounds as runtime arguments.
- Mark live route counts and workspace bounds with `do_not_specialize`.
- Add regression checks for both constexpr membership and Triton's `do_not_specialize` metadata.

The expert count, route block size, and binary-search geometry remain compile-time constants. Route ordering, output layout, and workspace sizing are unchanged.

## Validation

- Four focused regression cases passed in the CUDA image: one verifies that workspace capacities are not constexpr arguments; three verify the non-specialized runtime arguments on the count, post-prefix, and sort kernels.
- Manual GPU smoke test used four RTX PRO 6000 Blackwell GPUs with tensor parallelism and decode-context parallelism both set to four. A fresh Triton cache warmed both supported `topk_ids` pointer-alignment layouts on every worker.
- The serving correctness gate passed deterministic short answers, strict structured output, and three-key retrieval from a 32,933-token prompt.
- vLLM's verbose JIT monitor emitted no inference-time compilation warning for the three changed route-pack kernels.

### DIFF
--- a/b12x/moe/_shared/kernels/w4a16/route_pack.py
+++ b/b12x/moe/_shared/kernels/w4a16/route_pack.py
@@ -19,7 +19,7 @@ _SMALL_PREFIX_MAX_ROUTE_BLOCKS = 512
 _FAST_COUNT_BLOCK_T = 1024
 
 
-@triton.jit
+@triton.jit(do_not_specialize=["live_numel"])
 def _w4a16_route_count_kernel(
     topk_ids,
     expert_map,
@@ -122,7 +122,9 @@ def _workspace_slice(
     return tensor[:elements]
 
 
-@triton.jit
+@triton.jit(
+    do_not_specialize=["live_numel", "max_packed_routes", "max_route_blocks"]
+)
 def _pack_topk_routes_post_prefix_kernel(
     packed_route_indices,
     block_expert_ids,
@@ -130,8 +132,8 @@ def _pack_topk_routes_post_prefix_kernel(
     live_numel,
     BLOCK_SIZE: tl.constexpr,
     NUM_EXPERTS: tl.constexpr,
-    MAX_PACKED_ROUTES: tl.constexpr,
-    MAX_ROUTE_BLOCKS: tl.constexpr,
+    max_packed_routes,
+    max_route_blocks,
     BLOCK_T: tl.constexpr,
     SEARCH_STEPS: tl.constexpr,
 ):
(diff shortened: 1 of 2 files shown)


malaiwah:
@coderabbitai review

malaiwah:
The two pre-merge errors in the earlier incremental review were addressed in head `64db087`:

- the commit/PR is now classified as a correctness fix rather than a throughput claim;
- the validation prose defines the four-GPU tensor/decode-context topology, pointer-alignment layouts, and each correctness workload;
- the concurrency/throughput language was removed.

The code diff did not change, so CodeRabbit's explicit re-review command returned `No files to review`; the current CodeRabbit status context is successful and GitHub reports the PR cleanly mergeable. I left the docstring-coverage warning alone rather than add weightless docstrings to pre-existing internal kernels and test functions.

malaiwah:
Additional full-GLM serving evidence from AIBeast (my own run, not independent replication): https://github.com/malaiwah/glm52-exl3-vast/pull/59

Four RTX PRO 6000 Blackwell Workstation96GB GPUs,375W, TP4/DCP4, full GLM-5.3 EXL3 3.42bpw, native probabilistic MTP3, scheduler3072, sequence limit12, capture48, context520192. Exact old Gilded-r34-derived parent and source hashes are in the linked recipe.

I ported #226/#256/#257 together onto the installed source, retaining the old linear post-prefix implementation rather than importing newer whole modules. Real CUDA checks covered adequate fixed arenas, guards/sentinels, captured replay, and actual GLM E256/topk8/cap3072 tails2049/2303/3071/3072. The separate two-rank owning-device teardown/rollback test also passed.

The B12X-only serving arm completed48 synthetic requests plus three fresh64K prefill/decode-overlap trials with no engine failure signatures. No inference-time JIT warning for the three #256 route kernels was observed; other indexer/prefill/sampler first-use warnings remained. This is not a claim that all inference JIT disappeared.

No isolated throughput percentage is justified: live clients intermingled, and the fresh original-code return control showed the initial C1 difference was confounded. The final selected B12X+measured-fairness/2048-arena composition separately passed500K retrieval, C8/C12 and a30-minute60/60-health-sample soak. Those final numbers belong to the composition, not #256 alone.

Pinned evidence/source: https://github.com/malaiwah/glm52-exl3-vast/tree/2b6c75eeff5375e5fe6ab23ec44a674902b46d9f/maintenance/glm53-optimization-20260908
Route GPU receipt: https://github.com/malaiwah/glm52-exl3-vast/blob/2b6c75eeff5375e5fe6ab23ec44a674902b46d9f/maintenance/glm53-optimization-20260908/evidence/route-glm-cuda.txt

One separate lesson: this deployment uses the V2 model runner inside the V1 engine; the V1 proposer helper is not the active native-MTP path. That correction is explicitly retained in the experiment record.

PR #257, opened 2026-08-29
fix(comm): bind PCIe teardown barriers to pool devices

malaiwah:
PCIe channel teardown passes the pool's CUDA ordinal to every coordinated barrier. Rollback and close therefore synchronize on the intended device even when another CUDA device is current. CPU barrier behavior, IPC unmap-before-free ordering, failure retention, and retry semantics are preserved.

Lifecycle regressions assert the CUDA device argument alongside cleanup ordering and idempotence. A two-rank GPU regression exercises production PCIe channels, eager reduction, mutated-input CUDA graph replay, transient-channel rollback, continued use of retained channels, and repeated close from a different current device. Graph replay preserves addresses and performs no allocator activity.

Validation on NVIDIA RTX PRO 6000 Blackwell Max-Q GPUs with PyTorch 2.12.0+cu130 and CUTLASS DSL 4.6.2:

- `python -m pytest -q tests/comm/test_pcie_oneshot.py tests/comm/test_pcie_ipc_lifecycle.py`: 110 passed.
- `B12X_RUN_PCIE_TEARDOWN_TEST=1 python -m pytest -q -s tests/comm/test_pcie_teardown_gpu.py`: passed on two GPUs, with two create/replay/rollback/close cycles per rank and three graph replays per cycle.
- Ruff passed on all four changed files.

Public interfaces are unchanged. These checks qualify the channel teardown path; they do not claim a model-serving or performance result.

## Summary

PCIe channel teardown barriers now synchronize on the intended CUDA device. CUDA pools pass the device index through a shared barrier helper, which resolves the current device when no index is set. CPU groups retain the existing barrier behavior.

The change prevents teardown and rollback synchronization from using an unintended CUDA device. It preserves compatibility for CPU groups and unindexed CUDA devices.

Regression tests verify all three teardown barriers on `cuda:3` and rollback synchronization. PCIe teardown tests, TP4 startup, and serving tests for short, structured-output, and long-context workloads passed.


malaiwah:
On the GPU-backed-test nitpick: I kept the regression at the unit boundary intentionally. A production `PCIeOneshotAllReducePool` teardown test requires a multi-process CUDA group, at least two GPUs, peer IPC support, and driver/topology assumptions that are not available on ordinary CI runners. The changed contract is the `dist.barrier(..., device_ids=[pool_device])` dispatch at each existing teardown phase, which this test observes directly without weakening any cleanup path.

The integration path was exercised separately on four RTX PRO 6000 GPUs with `B12X_PCIE_ONESHOT_DMA`; five graceful TP4 service shutdowns/restarts completed with the patched close path, and the subsequent production startup remained clean. I would avoid adding a topology-dependent integration test unless the maintainers have a dedicated multi-GPU marker/runner for it.

PR #274, opened 2026-09-01
perf(moe): gate the W4A16 stripe split-K to decode-sized launches

myshytf:
## Summary

`B12X_W4A16_SMALL_M_SPLITK=1` swaps the one-tile-per-CTA small-M schedule for the stripe split-K partition in both W4A16 kernel classes. On Kimi-K3 QSRT trellis experts (TP8, 2026-08-28) that is a decode win (MoE kernel 235.2 → 198.5 µs, −2.9 ms per step) but the flag also stripped the large-M prefill launches, where the whole-tile wave schedule is already grid-saturated and the stripe finalize is pure overhead: prefill regressed 902 → 782 tok/s.

This applies the stripe only when the compiled `size_m` is ≤ 16 (decode/verification shapes); prefill buckets (≥ 512 tokens) keep the whole-tile schedule. `size_m` is already part of the kernel cache key, so per-M variants compile independently.

Served in production since 2026-08-28 as a boot-time patch on the f25c8bdd-based tree; this branch applies the identical gate on `master` (both anchors present, module compiles). No kernel math changes.

## Summary

- Updates comments for the existing `size_m <= 16` stripe split-K gate in both W4A16 kernel classes.
- Keeps stripe split-K for decode-sized launches.
- Keeps the whole-tile schedule for larger launches to avoid cross-CTA reduction overhead.
- Makes no executable or kernel-math changes.

### DIFF
--- a/b12x/moe/_shared/kernels/w4a16/kernel.py
+++ b/b12x/moe/_shared/kernels/w4a16/kernel.py
@@ -1015,9 +1015,9 @@ class W4A16GemmKernel:
         # Small-M stripe split-K: opt out of the one-tile-per-CTA fast path
         # so decode-heavy small-M phases spread each mn-tile's K range across
         # multiple CTAs (existing tail scheduling plus cross-CTA finalize).
-        # Stripe only decode-sized launches;
-        # large-M prefill keeps the whole-tile wave schedule (stripe
-        # finalize regressed prefill 902->782 tok/s on 2026-08-28).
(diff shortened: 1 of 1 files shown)


myshytf:
Review follow-up (b6e11f5d):
- applied: both split-K gate comments now state the scheduling invariant (decode-sized launches stripe; larger M keeps the whole-tile wave schedule and would pay a cross-CTA finalize per tile) instead of the dated throughput figures.

PR #281, opened 2026-09-01
perf(comm): CUDA-graph replay for eager PCIe DMA ring all-reduces

myshytf:
## What

`B12X_PCIE_DMA_GRAPH_REPLAY=1` makes `PCIeDmaAllReduce.all_reduce` capture each eligible shape (>= `B12X_PCIE_DMA_GRAPH_REPLAY_MIN_BYTES`, default 8 MiB) into a CUDA graph over static input/output buffers on the shape's second call, and replay it afterwards (copy in, replay, copy out). One-off shapes stay eager; an LRU of `B12X_PCIE_DMA_GRAPH_REPLAY_MAX_ENTRIES` (default 4) bounds the cache. The static buffers of all entries come from one arena reserved when the ring is constructed (`max_entries x 2 x max_bytes`), so a capture during serving allocates nothing and cannot fail for lack of device memory — there is no request-time fallback; a deployment that cannot afford the arena fails at start-up. Calls made while an enclosing stream capture is active keep recording the eager sequence into that graph, as before. Default off.

## Why

An eager ring call issues ~250 CUDA API calls from Python (per-piece copies, flag kernels, adds, cross-stream events). For prefill-size tensors this host time exceeds the wire time: on a Kimi-K3 TP8 / PCIe Gen4 x16 target a 22 MiB call costs ~2.4 ms of worker-thread time against ~1.4 ms on the wire, and the ring driver is 49% of the worker thread's CPU time during a 1,536-token chunk (py-spy, 100 Hz). Replay removes that host cost without touching any kernel; the kernels and their order are unchanged, so results are bit-identical.

## Validation

- `tests/comm/test_pcie_dma_gpu.py::test_pcie_dma_all_reduce_graph_replay_matches_eager` (gated by `B12X_RUN_PCIE_DMA_TEST=1`): eager vs replayed outputs `torch.equal` across three inputs of one shape, no allocator growth at capture beyond the returned output, eviction at `max_entries=1` with a second shape, and the enclosing-capture path. Written with this change; not yet executed here because the only multi-GPU host is serving (to be run in the next maintenance window).
- Serving qualification on the Kimi-K3 target (TP8/DCP8, 1,536-token chunks, 2026-09-02, lazy-buffer revision): replay active (py-spy: eager driver 0% of worker time, `_all_reduce_replayed` 15%), greedy / JSON-schema / 11.8k-token lookup outputs correct, no CUDA errors, decode unchanged (c1 43.6-51.0 tok/s); prefill wall time unchanged (8k 1,492-1,499 tok/s, cold 1,536 tokens 0.877 s) because that target's chunk is GPU-bound once the host cost is removed — the change buys host headroom, not wall time, there.

AI assistance was used for this change (Claude Code); the submitter reviewed every line.

## Summary

`PCIeDmaAllReduce` supports optional CUDA-graph replay for eager PCIe DMA ring all-reduces. When `B12X_PCIE_DMA_GRAPH_REPLAY=1`, eligible shapes are captured on the second call and replayed with static input and output buffers. Shapes below the 8 MiB default threshold and one-off calls remain eager.

The replay cache uses LRU eviction and holds up to four entries by default. Static buffers are reserved in one arena when the ring is created. Insufficient device memory fails startup. Calls inside an enclosing CUDA graph capture continue to record the eager sequence.

The feature is disabled by default. Existing eager kernels and ordering remain unchanged. The README documents the replay controls and cache behavior.

## Validation

A gated multi-GPU test checks eager and replay output equality, capture timing, buffer allocation, LRU eviction, reshaping, and enclosing stream capture. The test has not run because a suitable multi-GPU host was unavailable.

Kimi-K3 serving qualification confirmed replay activity, correct outputs, no CUDA errors, and lower eager-driver worker time. Prefill wall time was unchanged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,25 @@ Compressed transport requires BF16 input and a per-rank shard divisible by
 B12X_PCIE_DMA_FP8=i8_ring python -m your_server
 ```
 
+### Graph replay of eager ring all-reduces
+
+An eager `PCIeDmaAllReduce.all_reduce` call issues roughly 250 CUDA API calls
+from Python (per-piece copies, flag kernels, adds, cross-stream events). On a
+prefill-size tensor that host time (about 2.4 ms per 22 MiB call on a
+desktop-class host) exceeds the wire time, so a Python-driven serving loop
+becomes launch-bound. `B12X_PCIE_DMA_GRAPH_REPLAY=1` captures each eligible
+shape once into a CUDA graph over static input/output buffers (on the shape's
+second call, so one-off sizes stay eager) and replays it afterwards with a
+copy in and a copy out. The kernels and their order are unchanged, so the
+result is bit-identical to the eager call; calls made while an enclosing CUDA
+graph is being captured keep recording the eager sequence into that graph.
+
+| Variable | Default | Meaning |
+|---|---|---|
+| `B12X_PCIE_DMA_GRAPH_REPLAY` | `0` | `1` enables replay for eager calls |
+| `B12X_PCIE_DMA_GRAPH_REPLAY_MIN_BYTES` | `8388608` | Smallest tensor size that is captured |
+| `B12X_PCIE_DMA_GRAPH_REPLAY_MAX_ENTRIES` | `4` | Cached shapes; least recently used is dropped first. The static buffers for all entries (`2 x max_bytes` each) are reserved when the ring is constructed, so serving never allocates for a capture |
+
 Compilation happens lazily per shape/config and is cached. For serving, warm
 up the shapes you need, then freeze:
 
--- a/README.md
+++ b/README.md
@@ -154,15 +154,19 @@ B12X_PCIE_DMA_FP8=i8_ring python -m your_server
 ### Graph replay of eager ring all-reduces
 
 An eager `PCIeDmaAllReduce.all_reduce` call issues roughly 250 CUDA API calls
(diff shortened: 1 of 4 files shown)


myshytf:
Review follow-up (a56bb9ca):
- applied: the replay path enforces the same `out` contract as the eager path (shape/dtype/device/contiguity) before copying out of the static buffer.
- applied: evicting a cache entry drops the shape's warm-up count, so a shape returning after eviction runs one eager call before it is captured again (no capture on every call for cyclic shape sets).
- applied: a capture that raises returns its static-buffer slot; slots are owned by entries only once the capture completed.
- applied: README and the replay docstring state bit-identity as a property of the recorded kernel sequence verified by `tests/comm/test_pcie_dma_gpu.py`, and no longer quote an unrecorded timing figure.
- not applied: "avoid a dynamic output allocation on replay" — with `out=None` the replay path allocates exactly as the eager out-of-place collective does; callers that need stable storage pass `out`, on both paths. Replay does not add an allocation.
- already resolved earlier: replay stays enabled after a capture failure (the static arena is reserved at construction; there is no OOM disable path any more).
- new `tests/comm/test_pcie_dma_replay_bookkeeping.py` (one GPU, stubbed collective) covers the three invariants; the two-rank ring test was not run here (needs two free GPUs).

PR #283, opened 2026-09-02
feat(comm): add a lossless bf16 PCIe two-shot all-reduce

MadeBy561:
## Purpose

Add `PCIeTwoShotBF16`, a lossless bf16 two-shot all-reduce for TP decode payloads above the one-shot ceiling (tens of KB) and below the DMA ring floor (MB), where the NCCL ring is the incumbent today.

## Behavior

- Pull mode: each rank stages its bf16 payload in its IPC slab; after a per-CTA flag barrier every rank pulls the peers' packs of its quarter and accumulates them in fp32 in a fixed rank order; after a second barrier the reduced quarters are pulled into place. Two launches per all-reduce, double-buffered staging slots with device-side slot selection, so graphs captured from one instance replay correctly.
- The result is rounded to bf16 once, so every element lies within one bf16 rounding of the exact fp32 sum of the inputs. The bf16 NCCL ring rounds after every hop. Outputs are deterministic across calls and across graph replays.
- `accepts()` requires a contiguous bf16 tensor whose element count is a multiple of `row_elems * world_size` and at most `max_rows` rows; world sizes 2, 4 and 8. The fp8 two-shot (`pcie_twoshot.py`) is untouched.
- Graph capture follows the existing two-shot contract: `runtime.capture()` around `torch.cuda.graph`, and replays of graphs captured from one instance stay serialized.

## Validation

Correctness, four RTX PRO 6000 Blackwell (PCIe), run inside the GLM-5.3 serving image:

```text
python -m torch.distributed.run --nproc-per-node=4 tests/comm/test_pcie_twoshot_bf16.py
pcie_twoshot_bf16 correctness OK (4 ranks, rows (8, 16, 32, 64, 96, 128, 192, 256))
```

For every row count and three staging-slot rotations the test checks the one-bf16-rounding bound against the exact fp32 sum of the all-gathered inputs, that the two-shot error never exceeds the NCCL ring's, determinism across repeated calls, `accepts()` rejections (dtype, row width, row count, capacity), and that a captured graph replays bitwise equal to the eager result three times. The device-free layout test runs under pytest.

Timing (graph replay, 4096-element bf16 rows, 512 threads, two-shot vs NCCL ring, microseconds):

| Payload | Two-shot | NCCL ring |
| --- | ---: | ---: |
| 128 KB | 15.0 | 16.7 |
| 256 KB | 20.3 | 25.6 |
| 384 KB | 26.3 | 33.5 |
| 512 KB | 32.8 | 41.7 |
| 768 KB | 44.5 | 54.5 |
| 1 MB | 63.4 | 61.3 |
| 2 MB | 99.8 | 98.3 |

Both reach the same ~15 GB/s per-GPU fabric ceiling at 1 MB and above, so the useful window ends at 768 KB. The earlier push-mode port of the fp8 structure measured 1.3 ms at 2 MB (remote posted writes) and was discarded in favour of this pull design.

Precision: measured maximum error against the exact fp32 sum 0.0625 at row magnitude ~16 (one bf16 half-ulp) versus 0.14 to 0.17 for the NCCL bf16 ring on the same inputs.

Serving impact (GLM-5.3-Flash NVFP4 target, MXFP8 DFlash2 K7 draft, TP4, 98 decode all-reduces per verifier step, with the vLLM dispatch that routes 84 KB to 768 KB payloads here): the all-reduce drops from 41.7 to 32.8 us at 512 KB (concurrency 8) and from 54.5 to 44.5 us at 768 KB (concurrency 12), about 1 ms of a 31 to 36 ms step; measured C8 259.9 to 267.7 verifier steps/s (+3.0%, 620 to 659 output tok/s) and C12 307 to 315 (+2.6%, 788 to 801 tok/s) from this route alone, +7.2% and +20.8% together with the exact 96/192 graph sizes and the SM120 mHC split policy it shipped with. C1 (64 KB, one-shot) and C30 (1.97 MB, ring) payloads are outside the window and unchanged.

The vLLM dispatch is a separate pull request on `local-inference-lab/vllm`.

### DIFF
--- a/b12x/comm/pcie/_twoshot_bf16_cute.py
+++ b/b12x/comm/pcie/_twoshot_bf16_cute.py
@@ -1,0 +1,1059 @@
+"""CuTeDSL kernels for lossless BF16 PCIe two-shot collectives.
+
+Structure mirrors :mod:`_twoshot_cute` (the fp8-transport variant): phase one
+pushes this rank's shard into every peer's IPC staging slot (posted PCIe
+writes), a per-CTA flag barrier follows, and phase two either reduces the
+local quarter (reduce_scatter) or copies the staged shards into place
+(all_gather).  Payload packs are 16 bytes = 8 bf16 values; the reduction
+accumulates in fp32 in a fixed rank order (local rank first, then
+``(local + i) % world`` for ``i = 1..world-1``) and rounds once to bf16, so a
+given rank's output is deterministic across runs.
+"""
+
+from __future__ import annotations
+
+import functools
+from collections.abc import Callable, Sequence
+
+import cuda.bindings.driver as cuda
+import cutlass
+import cutlass.cute as cute
+from cutlass import Float32, Int32, Int64, Uint32
+
+from b12x._lib.compiler import KernelCompileSpec
+from b12x._lib.compiler import compile as b12x_compile
+from b12x._lib.intrinsics import ld_global_nc_v4_u32, ld_global_v4_u32, st_global_v4_u32
+from b12x._lib.runtime_control import raise_if_kernel_resolution_frozen
+from b12x._lib.utils import current_cuda_stream, make_ptr
+
+from ._cute_intrinsics import (
+    graph_epoch_arrive_serialized,
+    ld_relaxed_gpu_u32,
+    pack_f32x2_to_bf16x2,
+    unpack_bf16x2,
+)
+from ._twoshot_cute import (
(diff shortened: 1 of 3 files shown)


voipmonitor:
Superseded by #290. The replacement preserves MadeBy561 as the implementation author and adds the verified tensor, alignment, capacity, and NCCL ring contract checks. Closing this pull request keeps a single merge candidate.

PR #279, opened 2026-09-01
perf(runtime): add PDL between PCIe all-reduce and MHC

voipmonitor:
## Behavior

Allow programmatic dependent launch between the PCIe one-shot all-reduce producer and the MHC partial consumer. `B12X_PCIE_ONESHOT_PDL=1` releases dependent work only after all publication and epoch-rollover barriers complete; `B12X_MHC_PDL=1` enables the consumer launch/wait pair. Both options remain disabled by default.

The numerical operation, caller-owned scratch and peer synchronization contract are unchanged. CUDA graph capture remains supported. The branch includes B12X master `00b69ac22e21`; it does not require a generated policy change or the separate M8 route-packing PR.

## Validation

Status: **implemented; synchronization contracts qualified**.

- Six PCIe torture tests pass on two RTX PRO 6000 Workstation GPUs with PDL enabled: eager execution, CUDA graphs, multiple streams and epoch rollover.
- Invocation: `B12X_RUN_PCIE_ONESHOT_TORTURE=1 B12X_PCIE_ONESHOT_PDL=1 B12X_MHC_PDL=1 pytest tests/comm/test_pcie_oneshot_torture.py`.
- Eight specialized MHC graph cases and the exact GLM M8 graph oracle pass in the composed B12X source.
- The separate fused MHC post/pre oracle exceeds its existing tolerance at the same two BF16 elements on master, published R34 and this composition: maximum absolute error 0.0078125 versus tolerance 0.004. No tolerance was weakened, and that reference discrepancy is not claimed fixed.

The [stock-clock serving evidence](https://github.com/voipmonitor/b12x/blob/perf/pcie-mhc-pdl-20260901/docs/evidence/glm53_pcie_mhc_dependent_launch_20260901.md) records the qualified 76.653 → 80.074 verifier steps/s comparison (+4.46%). That measurement belongs to the revisions in the report, not an unmeasured claim about this master merge.

MadeBy561's implementation authorship is preserved. Integration and validation used OpenAI Codex assistance.

PR #270, opened 2026-09-01
perf(mhc): profile GLM-5.3 decode partial grouping

voipmonitor:
## Behavior

MHC planning selects a typed decode partial-group schedule. For the measured native plan with 128-token capacity, hidden size 4096 and split-K 64, the selected schedule groups 13 partials per thread block. Other capacities retain their default schedule; medium-prefill projection geometry is unchanged.

`MhcConfig` requires `decode_partials_schedule`; its serialized config schema is 3. Overrides can use `default`. Bindings continue to use caller-owned scratch, and the policy is resolved at plan time rather than during graph replay.

## Integration and attribution

This PR targets `master` directly, including the merged recurrent-policy work from #269. It replaces #264. MadeBy561 retains authorship of the policy, generator, launch, and test commit. Martin Vít retains authorship of the fake-dispatch test correction and Max-Q profile entry.

Only the `norm.mhc` component changes in the three generated profile archives. All other components and the baseline device-target ordering are preserved. The MHC components exactly match the GLM community qualification source.

## Validation

Status: implemented; CPU policy and dispatch contracts qualified. An isolated end-to-end gain is not established.

- **66 CPU policy, scratch-binding, and fake-dispatch tests passed** against the master-based source; Ruff and whitespace checks passed.
- Component audit: GB10, RTX PRO 6000 Workstation/Server, and Max-Q profiles retain every unrelated master component and match the qualified MHC component.
- The component-transplant utility passed six identity, preservation, and uniqueness tests.
- Previously measured 32-token kernel latency was 17.37 → 16.23 microseconds (6.6% lower). The master-based source has not received an independent repetition of that isolated GPU measurement.

## AI assistance

OpenAI Codex assisted with port construction and validation under Martin Vít's direction.

## Summary

- Added configurable decode partial grouping for mHC post-pre partial launches.
- Added `hidden4096_m128_v1` for native plans with `max_tokens=128`, `hidden_size=4096`, and `split_k=64`.
- Preserved the default schedule for other configurations.
- Added schedule validation and propagated the value through planning, runtime dispatch, and fake dispatch.
- Updated `MhcConfig` serialization to schema version 3 and added `default` overrides.
- Updated generated profiles and tuner candidates.

## Validation

- 66 CPU policy, scratch-binding, and fake-dispatch tests passed.
- Six profile-transplant tests passed.
- Ruff and whitespace checks passed.
- Full-stack validation passed on four RTX PRO 6000 GPUs.
[...]

PR #284, opened 2026-09-02
perf(mhc): select the SM120 decode source split and prefill crossover

MadeBy561:
## Purpose

Select the faster mHC decode schedule for RTX PRO 6000 Blackwell (SM120) on GLM-5.3 (hidden size 4096): the four-way source split of the post/pre partial kernel from 32 padded rows up, and a prefill crossover of 192 rows instead of 96, both measured on the GLM-5.3 TP4 decode graphs. This is the policy the R22 serving stack has run since 2026-09-01; it is filed so the maintainer can weigh it with the numerics note below.

## Behavior

- `_selected_post_pre_decode_split_n` returns the (4, 6) split for SM120 decode graphs with at least 32 rows and the unsplit kernel below; the GB10 (12, 1) policy and the `B12X_MHC_DECODE_SPLITS` / `B12X_MHC_DECODE_TILE_N` overrides are unchanged.
- The default `B12X_MHC_PREFILL_MIN_TOKENS` becomes 192 on SM120 (96 elsewhere); the environment override still wins. The value is resolved once per device and cached, so it is graph-capture safe.
- Nothing changes for other architectures or hidden sizes.

## Numerics (read this first)

The residual update (`out`) is bitwise identical to the unsplit kernel. The 25 fn partials and 10 Gram pairs are reduced over four source splits in the finalize, so the gate/mix values (`post`, `comb`) differ from the unsplit kernel by fp32 rounding (1 to 2 ulp measured at 32/64/128 rows) and `y` by at most one bf16 ulp. The new test pins that contract (`out` equal, `y` within one bf16 ulp, `post`/`comb` within a few fp32 ulp, deterministic). This is a summation-order change, not a precision loss, but it is not bit-exact; the C30 greedy gates of the serving stack use 240-row graphs, which take the prefill kernels, so they do not exercise the 32 to 191-row window this policy applies to (concurrency 4 to 23 at eight tokens per request).

## Compatibility

Merge simulation against #269, #270 and #279 (which touch the same files) completes without conflicts. No profile archives change.

## Validation

Tests inside the GLM-5.3 serving image on one RTX PRO 6000 Blackwell:

```text
python -m pytest tests/norm/test_mhc_sm120_decode_policy.py
16 passed
```

Policy selection for SM120, GB10 and other capabilities, the hidden-size guard, the override precedence, the crossover default, and the split-vs-unsplit contract at 32, 64 and 128 rows. The pre-existing `tests/norm/test_mhc.py` reference-tolerance failures at 3 and 8 rows and its graph-capture case fail identically on unmodified `master` and on the shipped image in this environment; this change does not alter them.

Kernel time, decode post/pre partial + finalize pair, unsplit vs split-4 (microseconds, same inputs): 32 rows 20.2 vs 16.4, 48 rows 28.1 vs 22.6, 64 rows 35.1 vs 26.7, 96 rows 43.0 vs 33.3, 128 rows 44.8 vs 41.0. Against the block-M prefill kernel: 96 rows 33.6 vs 43.0, 128 rows 41 vs 45, 192 rows 57 vs 47.1 (hence the crossover at 192).

Serving (GLM-5.3-Flash NVFP4 target, MXFP8 DFlash2 K7 draft, TP4, llm_decode_bench ctx0 30-second cells): the C12 step's mHC time fell from 61 us x 86 calls to 26 us x 86 calls (about 3 ms of a 42 ms step). Measured together with the exact 96/192 CUDA-graph sizes (`--cudagraph-capture-sizes 8 16 32 64 96 128 192 256`, a compose-level change worth another 4 ms per C12 step because padded rows no longer route to experts) and the bf16 two-shot all-reduce (#283): C8 617 to 659 tok/s (+7.2% verifier steps/s), C12 638 to 801 tok/s (+20.8%), C24 967 to 1108 tok/s (+15.2%), C1 and C32 unchanged (their graphs are outside the window).

## Resulting behavior

- On SM120, GLM-5.3 mHC decode with hidden size 4096 uses source split `(4, 6)` when the padded row count is at least 32.
- On SM120, prefill begins at 192 padded rows instead of 96.
- Other architectures and hidden sizes retain existing policies.
- Environment-variable overrides retain precedence.
- The prefill crossover value is cached per device for graph-capture safety.

## Compatibility and numerical impact

- The residual update remains bitwise identical.
- Reduction-order changes can alter `post` and `comb` by a few fp32 ulp.
- `y` can differ by at most one bf16 ulp.
- GB10 behavior remains unchanged.

## Validation

- Sixteen policy tests pass on an RTX PRO 6000 Blackwell.
- CUDA integration tests verify deterministic split and unsplit decode results within the defined tolerances.
- Benchmarks show lower kernel times for the selected split and higher serving throughput at C8, C12, and C24.

### DIFF
--- a/b12x/norm/mhc/_impl.py
+++ b/b12x/norm/mhc/_impl.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import functools
 import os
 from collections.abc import Mapping, Sequence
 from dataclasses import dataclass
@@ -875,6 +876,42 @@ def _b12x_mhc_pre_impl(
     )
 
 
+
+_DEFAULT_MHC_PREFILL_MIN_TOKENS = 96
+_SM120_MHC_PREFILL_MIN_TOKENS = 192
+
+
+def _mhc_prefill_min_tokens_for_capability(
+    compute_capability: tuple[int, int] | None,
+) -> int:
+    """Padded row count from which the prefill mHC kernels replace the decode
+    partial kernel when no override is configured.
+
+    On SM120 (RTX PRO 6000 Blackwell) the four-way decode source split is
+    faster than the block-M prefill kernel up to 128 padded rows (96 rows:
+    33.6 vs 43.0 us; 128 rows: 41 vs 45 us) and slower from 192 rows
+    (57 vs 47 us, measured 2026-09-01 on GLM-5.3 hidden size 4096), so the
+    crossover moves from 96 to 192 rows there.
+    """
+    if compute_capability == (12, 0):
+        return _SM120_MHC_PREFILL_MIN_TOKENS
+    return _DEFAULT_MHC_PREFILL_MIN_TOKENS
+
+
+@functools.lru_cache(maxsize=None)
+def _cuda_prefill_min_tokens(device_index: int) -> int:
+    return _mhc_prefill_min_tokens_for_capability(
+        tuple(torch.cuda.get_device_capability(device_index))
+    )
+
+
+def _default_mhc_prefill_min_tokens(device: torch.device) -> int:
+    if device.type != "cuda":
+        return _DEFAULT_MHC_PREFILL_MIN_TOKENS
+    index = device.index if device.index is not None else torch.cuda.current_device()
+    return _cuda_prefill_min_tokens(int(index))
+
+
 def _b12x_mhc_post_pre_impl(
     x: torch.Tensor,
     residual: torch.Tensor,
@@ -1161,7 +1198,10 @@ def _b12x_mhc_post_pre_impl(
             )
 
         prefill_min_tokens = int(
-            os.environ.get("B12X_MHC_PREFILL_MIN_TOKENS", "96")
+            os.environ.get(
+                "B12X_MHC_PREFILL_MIN_TOKENS",
+                str(_default_mhc_prefill_min_tokens(residual.device)),
+            )
         )
         if planned_config is not None:
             tf32_enabled = planned_config.backend == "tf32_tma"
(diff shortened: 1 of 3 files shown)


lukealonso:
This isn't the primary mechanism for policy selection anymore. We have an autotuner system now. If these parameters aren't yet part of the autotuner, that's the correct fix.

voipmonitor:
## Independent validation

The SM120 mHC policy suite passed all 16 tests on an NVIDIA RTX PRO 6000
Blackwell Workstation Edition GPU. The checks cover architecture and hidden-
size selection, override precedence, the 192-row prefill crossover,
determinism, bitwise-equal residual updates, FP32 reduction-order tolerance,
and the one-BF16-ulp output contract.

Matched serving A/B used four stock-clock RTX PRO 6000 Blackwell Workstation
Edition GPUs, TP4/DCP1, an NVFP4 target, an MXFP8 DFlash2 K7 draft, 15 seconds
of warmup, and three 30-second context-zero cells per concurrency. The control
used the pre-existing unsplit/crossover policy; the candidate used this PR's
SM120 policy:

| Concurrency | Existing policy steps/s | PR policy steps/s | Change | Existing policy tok/s | PR policy tok/s | Change |
|---|---:|---:|---:|---:|---:|---:|
| C8 | 284.71 | 286.88 | +0.76% | 719.00 | 752.64 | +4.68% |
| C12 | 325.74 | 350.60 | +7.63% | 850.56 | 905.57 | +6.47% |

Target steps/s is the primary execution metric because DFlash output tok/s also
depends on stochastic acceptance. C1 is below the split threshold, and C24 is
at the prefill-kernel side of the crossover. The result confirms a material C12
serving gain and the declared numerical contract.

lukealonso:
The current-master port is #310. It preserves MadeBy561 as the commit author, moves every post/pre compile-time schedule into the typed mHC policy, and embeds a qualified RTX PRO 6000 profile. The replacement remains draft until the GB10 schema-4 profile is generated and qualified.

lukealonso:
Superseded by #310, which preserves the original authorship and carries the SM120 mHC source-split implementation with plan-time policy integration and an in-layer RTX PRO 6000 profile.

PR #285, opened 2026-09-02
perf(gdn): hoist the KDA decode operand loads before the recurrence

MadeBy561:
## Purpose

Take the eight dependent global-load round trips out of the serial recurrence of the packed sequential KDA decode kernel: with the 16-wide recurrent value tile, every token's q, k, v, beta, gate and destination slot are loaded before the recurrence.

## Behavior

- Applies when `BLOCK_V <= 16` (the block-V16 plan profiled for GLM-5.3 on RTX PRO 6000 Blackwell, #269, already in the shipped image). With the 32-wide tile the hoisted variant spills and is slower (20.6 to 50 us at one request in an ungated build), so the generic loop stays for `BLOCK_V > 16`.
- The per-token arithmetic sequence is the generic loop's (bf16 operands convert to fp32 at use, same order), so outputs and recurrent states are bitwise identical.

## Overlap

The change sits in `_packed_sequential_kda_decode_kernel`, which #253 (ours, parked after a multi-request acceptance regression) also edits; the two conflict textually. If #253 proceeds it will be rebased on this; otherwise it should be closed.

## Validation

Bitwise: output and recurrent state dumps from the GLM-5.3 TP4 decode shape at 1, 8 and 32 requests (eight tokens each) are `torch.equal` between the generic and hoisted kernels, on both the shipped image tree (block-V16) and unmodified `master` (block-V32, generic path).

Tests inside the GLM-5.3 serving image on one RTX PRO 6000 Blackwell:

```text
python -m pytest tests/sequence/test_gdn_decode_kda.py
17 passed   (master tree)
```

On the shipped image tree the suite passes except `test_kda_rmsnorm_kernel_avoids_intermediate_bf16_rounding`, which fails identically without this change.

Kernel time (recurrence + norm per call, CUDA-graph replay, GLM-5.3 TP4 shape, 16 heads):

| Requests | Generic | Hoisted (block-V16) |
| --- | ---: | ---: |
| 1 | 14.4 us | 12.4 us |
| 8 | 28.7 us | 28.7 us |
| 32 | 182 us | 182 us |

Serving: 31 KDA layers per step, about 0.07 ms per verifier step at concurrency 1 (about 0.7%); the R22 stack that carries it measured 92.3/91.8/92.6 verifier steps/s single cells against 90.5/89.9/89.5 for the same stack without it, within the run-to-run band as expected for its size.

### DIFF
--- a/b12x/sequence/gdn_decode/_kernels.py
+++ b/b12x/sequence/gdn_decode/_kernels.py
@@ -256,85 +256,311 @@ def _packed_sequential_kda_decode_kernel(
         tl.float32
     )
 
-    for relative_token in range(STATE_INDEX_COLUMNS):
-        if (relative_token < state_index_columns) & (relative_token < (end - start)):
-            token = start + relative_token
-            token_i64 = token.to(tl.int64)
-            mixed_base = token_i64 * stride_mixed_token
-            q_offsets = key_head * KEY_HEAD_DIM + key_cols
-            k_offsets = KEY_HEADS * KEY_HEAD_DIM + q_offsets
-            v_offsets = (
+    # The hoisted variant below pays for its extra live registers: with the
+    # 16-wide recurrent value tile it is faster (GLM-5.3 TP4 decode on RTX PRO
+    # 6000 Blackwell: 14.5 -> 12.4 us per call at one request, unchanged at
+    # 8-32 requests), with the 32-wide tile it spills and is slower, so the
+    # generic loop stays for BLOCK_V > 16.
+    if BLOCK_V <= 16:
+        # Every per-token operand (q, k, v, beta, gate, destination slot) is
+        # loaded before the serial recurrence so the eight dependent global-load
+        # round trips leave the critical path.  The per-token arithmetic sequence
+        # is the generic loop's (bf16 operands convert to fp32 at use), so
+        # outputs and states are bitwise identical to it.
+        A_log_value = tl.load(A_log + value_head).to(tl.float32)
+        gate_bias = tl.load(
+            dt_bias
+            + value_head.to(tl.int64) * stride_dt_bias_head
+            + key_cols.to(tl.int64),
+            mask=key_mask,
+            other=0.0,
+        ).to(tl.float32)
+        for t in tl.static_range(STATE_INDEX_COLUMNS):
+            ok_t = (t < state_index_columns) & (t < (end - start))
+            token_i64_t = (start + t).to(tl.int64)
+            mixed_base_t = token_i64_t * stride_mixed_token
+            q_offsets_t = key_head * KEY_HEAD_DIM + key_cols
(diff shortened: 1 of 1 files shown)


MadeBy561:
Post-merge serving A/B: this kernel's isolated gain does not hold end to end, and I recommend reverting it.

Interleaved fresh boots on the R22 serving stack (GLM-5.3-Flash NVFP4 target, MXFP8 DFlash2 K7 draft, TP4, four RTX PRO 6000 Blackwell), R21 = the same stack without this kernel, R22 = with it, sequence R21, R22, R21, R22; each arm: one-minute idle, two five-run Sieve coding sets, four-cell C1 ruler (llm_decode_bench ctx0, 30-second cells):

| Arm | C1 verifier steps/s (4 cells, median) | Sieve median tok/s (2 sets) | Sieve max |
| --- | ---: | ---: | ---: |
| R21 (generic loop) | 92.49/91.82/91.09/92.28 (92.05) | 509.6, 479.2 | 546.8, 521.5 |
| R22 (hoisted) | 89.77/90.49/90.34/90.38 (90.36) | 479.8, 471.6 | 530.3, 499.2 |
| R21 (generic loop) | 92.51/92.40/92.25/91.51 (92.33) | 504.6, 453.3 | 530.7, 471.8 |
| R22 (hoisted) | 91.43/89.72/89.96/89.67 (89.84) | 470.8, 467.4 | 482.2, 486.9 |

The hoisted kernel measures 12.4 vs 14.4 us per call in the isolated CUDA-graph microbenchmark, but the serving step is about 2% slower with it in both interleaved arms: the microbenchmark does not represent the kernel next to the concurrent streams of the real step. On current `master` the hoisted path is inert (the recurrent tile is 32 wide, so the generic loop runs); it activates once the block-V16 policy (#269) lands, which is when the regression would appear. Revert PR follows.

voipmonitor:
Independent serving A/B reproduces the regression under the configuration where the hoisted branch is active.

Conditions: GLM-5.3-Flash NVFP4 target, MXFP8 DFlash2 K7 draft, TP4, four RTX PRO 6000 Blackwell GPUs at stock clocks, identical vLLM and B12X packages except for `b12x/sequence/gdn_decode/_kernels.py`, separate warmed JIT caches, and persistent L2 reservation disabled. The embedded RTX PRO 6000 profile selects `recurrent_block_v=16`. The file on merged B12X `master` is byte-identical to the file tested for #285.

Three-run C1 medians:

- generic recurrent loop: 88.98 target steps/s
- hoisted operand loads: 86.99 target steps/s
- change: -2.24%

After reversing the boot order, the generic-loop arm measured 89.77 target steps/s, so the C1 result is not explained by launch order or a cold cache. The initial three-run C8 medians were 294.86 versus 289.35 target steps/s (-1.87%), although a later C8 cell was noisy; C1 is the stable regression signal.

This agrees with the interleaved result above. The isolated CUDA-graph kernel improvement does not survive the concurrent serving step, so I also recommend reverting #285.

PR #286, opened 2026-09-02
Revert "perf(gdn): hoist the KDA decode operand loads before the recurrence"

MadeBy561:
## Purpose

Revert #285 (`64b2b0f`). The hoisted KDA decode kernel wins its isolated CUDA-graph microbenchmark (12.4 vs 14.4 us per call, bitwise identical) but makes the serving step slower, so it should not stay in `master`.

## Evidence

Interleaved fresh boots on the GLM-5.3-Flash serving stack (NVFP4 target, MXFP8 DFlash2 K7 draft, TP4, four RTX PRO 6000 Blackwell). R21 is the stack without the kernel, R22 the same stack with it; the two composes differ only in that kernel file. Sequence R21, R22, R21, R22; each arm: one-minute idle, two five-run Sieve coding sets (4096-token cap), four-cell C1 ruler (llm_decode_bench ctx0, 30-second cells, ordinary sampling).

| Arm | C1 verifier steps/s (4 cells, median) | Sieve median tok/s (2 sets) | Sieve max |
| --- | ---: | ---: | ---: |
| R21 (generic loop) | 92.49/91.82/91.09/92.28 (92.05) | 509.6, 479.2 | 546.8, 521.5 |
| R22 (hoisted) | 89.77/90.49/90.34/90.38 (90.36) | 479.8, 471.6 | 530.3, 499.2 |
| R21 (generic loop) | 92.51/92.40/92.25/91.51 (92.33) | 504.6, 453.3 | 530.7, 471.8 |
| R22 (hoisted) | 91.43/89.72/89.96/89.67 (89.84) | 470.8, 467.4 | 482.2, 486.9 |

About 2% of the verifier step in both interleaved arms. The microbenchmark runs the kernel alone on an idle GPU; in the real step it runs next to the shared-expert, prefetch and all-reduce streams, where its extra live registers cost more than the removed load latency.

## Scope

On current `master` the hoisted path is inert: the recurrent value tile is 32 wide, so the generic loop runs. The regression would appear once the block-V16 policy (#269) lands. Reverting now keeps `master` equal to what has been qualified. The branch stays on the fork for reference.

Generated with Claude Code; the submitter ran the A/B on the listed hardware.

## Summary

- Restores the generic per-token recurrence loop in `_packed_sequential_kda_decode_kernel`.
- Loads operands inside the recurrence instead of using the hoisted register-cached path.
- Preserves kernel signatures, launch configuration, state updates, masking, normalization, decay, gating, and output behavior.
- Avoids the hoisted path's lower verifier and Sieve performance under concurrent-stream serving workloads.
- Isolates CUDA-graph microbenchmark gains are not retained.
- The change has no public API or compatibility impact.
- Serving benchmarks show higher C1 verifier medians and Sieve throughput with the generic loop.

### DIFF
--- a/b12x/sequence/gdn_decode/_kernels.py
+++ b/b12x/sequence/gdn_decode/_kernels.py
@@ -256,311 +256,85 @@ def _packed_sequential_kda_decode_kernel(
         tl.float32
     )
 
-    # The hoisted variant below pays for its extra live registers: with the
-    # 16-wide recurrent value tile it is faster (GLM-5.3 TP4 decode on RTX PRO
-    # 6000 Blackwell: 14.5 -> 12.4 us per call at one request, unchanged at
-    # 8-32 requests), with the 32-wide tile it spills and is slower, so the
-    # generic loop stays for BLOCK_V > 16.
-    if BLOCK_V <= 16:
-        # Every per-token operand (q, k, v, beta, gate, destination slot) is
-        # loaded before the serial recurrence so the eight dependent global-load
-        # round trips leave the critical path.  The per-token arithmetic sequence
-        # is the generic loop's (bf16 operands convert to fp32 at use), so
-        # outputs and states are bitwise identical to it.
-        A_log_value = tl.load(A_log + value_head).to(tl.float32)
-        gate_bias = tl.load(
-            dt_bias
-            + value_head.to(tl.int64) * stride_dt_bias_head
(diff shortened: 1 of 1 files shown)


voipmonitor:
Superseded by merged pull request #289, which restores the same generic KDA recurrent loop on B12X master. No additional source change from this pull request is required.

PR #287, opened 2026-09-02
perf(moe): direct SQG-XOR-Cheb-T12 rate table in shared memory for W4A16 trellis decode

myshytf:
## Summary

The fused W4A16 trellis MoE kernel decodes every weight by ranking its 16-bit window with a twelve-instruction integer bijection and then reading the 4 KiB modal T12 staircase staged in shared memory. On the one-CTA-per-SM decode grid the kernel is instruction-bound (RTX PRO 6000 / GB202: 188 persistent CTAs, 255 registers, 0.5–0.7 TB/s of 1.8 TB/s HBM), so the bijection dominates its time.

This change adds an opt-in decode path that stages the 64 KiB rate slice of the precomposed direct table (`byte(state, bits) = table[((bits - 2) << 16) | state]`, `sqg_xor_cheb_t12_direct_lut`) in shared memory and decodes each weight with one `ld.shared.u8` through a new intrinsic, `packed_decode_trellis_sqg_direct_lut_smem_to_e4m3x8`. The bytes are the same as the modal path, so the kernel output is bit-identical.

`B12X_SQG_XOR_CHEB_T12_DIRECT_SMEM=1` turns it on; off by default. The switch is part of the kernel cache key and is reported on the compile result (`sqg_xor_cheb_t12_direct_smem`), and launches pass the matching rate slice.

## Scope

- Uniform-rate `trellis_t256` kernels on a one-CTA-per-SM grid whose pipeline leaves 64 KiB under the per-block shared-memory limit. At 128x128 tiles that is the 2-bit pipeline: 34,944 B at route block 8 (100,480 B with the table under the 101,376 B SM120 limit) and 35,072 B at block 16.
- 3- and 4-bit pipelines (43,136 / 51,328 B) and wider route blocks (32: 51,712 B; 48: 68,352 B) keep the modal table and log one warning per footprint (`... direct table does not fit ...`).
- Multi-CTA grids keep the modal table (the grid is sized before the table is added; two CTAs cannot both hold 64 KiB).
- Pair-rate kernels keep the modal table.
- The mixed-rate kernels (`mixed_trellis.py`) stage the modal table only: their tiers are compiled with the new constructor argument `sqg_xor_cheb_t12_direct_smem=False`, and composing a direct-table tier is refused with a `ValueError`.

## Validation

RTX PRO 6000 (SM120), Kimi-K3 QSRT K2 coupled expert weights (`k2_coupled_h512_h128`, H=7168, I=384 per TP8 rank, 112 experts per rank, top-16):

| launch | modal (us) | direct (us) | output |
|---|---|---|---|
| decode, route block 8, m=8 (layers 1/50/92) | 145 | 99 (-32 %) | bit-identical, m=1/4/8 |
| decode, route block 8, m=4 | 99 | 68 | bit-identical |
| route block 16, m=256 | 1,941 | 1,296 | bit-identical |
| route block 16, m=1536 | 4,263 | 2,918 | bit-identical (m=256, 1536) |
| route block 48, m=1536 (does not fit) | 2,510 | 2,510 (fallback) | unchanged |

`tests/moe/test_w4a16_direct_table_smem.py` (9 tests, all pass with the switch on): bitwise equality against the modal path eagerly and under CUDA-graph capture at m=1/4/8; cache-key participation; 3/4-bit fallback with a single warning; multi-CTA grid skip; mixed-rate composition guard. W4A16 suites (`test_w4a16_{reference,mixed_trellis,packed_format,route_pack,tile_selection,e2e}.py`) with the switch on have the same failure set as without it.

The same change on the served production snapshot (46c84bab lineage) is branch `agent/kimi-k3-w4a16-direct-t12-table-20260902-served`.

## Summary

Adds an opt-in shared-memory decode path for uniform-rate W4A16 `trellis_t256` MoE kernels.

- `B12X_SQG_XOR_CHEB_T12_DIRECT_SMEM=1` stages a 64 KiB direct rate-table slice in shared memory.
- The new `packed_decode_trellis_sqg_direct_lut_smem_to_e4m3x8` intrinsic decodes weights with shared-byte lookups.
- The option is included in kernel cache keys and compile results.
- Launches pass the selected bitrate slice to the kernel.
- Multi-CTA, pair-rate, mixed-rate, unsupported-bitrate, and insufficient-shared-memory configurations use the existing modal table.
- Mixed-rate kernels reject direct-table tiers and compile their tiers with the modal table.
- The shared-memory path is compatible with eager and CUDA-graph execution.

The direct table reduces decode overhead by replacing modal-table decoding with direct lookups. Validation shows bit-identical output and improved RTX PRO 6000/SM120 performance, including 145 µs to 99 µs for route block 8 with `m=8`.

Nine SM120/SM121 tests cover eager and CUDA-graph correctness, cache keys, launch staging, fallback behavior, grid selection, and mixed-rate guards.

### DIFF
--- a/b12x/_lib/intrinsics.py
+++ b/b12x/_lib/intrinsics.py
@@ -7159,6 +7159,89 @@ def packed_decode_trellis_sqg_state_smem_to_e4m3x8(
 
 
 @dsl_user_op
+def packed_decode_trellis_sqg_direct_lut_smem_to_e4m3x8(
+    win_a,
+    win_b,
+    direct_lut_smem_addr,
+    bits: int = 3,
+    *,
+    loc=None,
+    ip=None,
+):
+    """Decode eight L16 windows through a direct E4M3 table in shared memory.
+
+    Shared-memory form of ``packed_decode_trellis_sqg_direct_lut_to_e4m3x8``:
+    the caller stages the 64 KiB rate slice of the direct table
+    (``byte(state) = table[state]``, the frozen XOR-Cheb rank map composed
+    with the modal T12 staircase, so lookups are bit-identical to the T12
+    decode) at a 32-bit shared byte offset. Per weight: one ``bfe``, one
+    ``add``, one ``ld.shared.u8`` and the pack, versus the twelve-instruction
+    bijection the T12 path evaluates before its lookup.
+    """
+    bits = int(bits)
+    if bits not in (2, 3, 4):
+        raise ValueError(
+            f"unsupported SQG-normal trellis bitrate {bits}; expected 2, 3, or 4"
+        )
+    extract_lines: list[str] = []
+    load_lines: list[str] = []
+    pack_lines: list[str] = []
+    for index in range(8):
+        source = "$3" if index < 4 else "$2"
+        shift = (3 - (index & 3)) * bits
+        target = "out0" if index < 4 else "out1"
+        byte_shift = 8 * (index & 3)
+        if shift:
+            extract_lines.append(f"bfe.u32 w{index}, {source}, {shift}, 16;")
+        else:
+            extract_lines.append(f"and.b32 w{index}, {source}, 0xffff;")
+        load_lines.append(
+            f"""
+                add.u32 addr{index}, w{index}, $4;
+                ld.shared.u8 w{index}, [addr{index}];
+            """
+        )
+        if byte_shift:
+            pack_lines.append(f"shl.b32 w{index}, w{index}, {byte_shift};")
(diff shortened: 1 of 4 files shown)


myshytf:
Review follow-up (379151eb):
- applied: the bitwise modal-vs-direct test requires a nonzero modal output before the equality checks.
- applied: the compiled-launch test records `shared_memory_bytes` and requires the direct-table launch to carry at least the 64 KiB table and to grow by the table minus the 4 KiB modal table it replaces (9 passed).

PR #288, opened 2026-09-02
perf(gemm): quantize small-M MXFP8 linear activations inside the GEMM

myshytf:
## Summary

The MXFP8 dense linear (ModelOpt/serialized MXFP8 weights with dynamic MXFP8 activations, `b12x.gemm.blockscaled` / `mxfp8_linear`) ran the standalone activation quantizer before every GEMM: two E8M0 scale-buffer fills, the row quantization kernel, then the dense GEMM. For up to eight tokens with BF16 activations, an unpadded K and an N that is a whole number of 64-wide tiles and at most `B12X_MXFP8_LINEAR_FUSED_QUANT_A_MAX_N` (default 4096), the linear now runs `dense_gemm_fused_quant_a` — the small-M GEMM that quantizes A in each CTA, which the block-FP8 linear already uses for the same token range. The in-CTA quantization computes the same UE8M0 scales and E4M3 values, so the output is bit-identical; three launches per projection disappear.

Wider N keeps the separate path: the fused GEMM re-quantizes A per N tile and has no split-K, and at N = 7168 it is slower than the quantizer plus the split-K GEMM (18.6 vs 15.7 µs on SM120). `B12X_MXFP8_LINEAR_FUSED_QUANT_A_MAX_N=0` disables the routing.

## Validation (RTX PRO 6000, SM120)

Kimi-K3 decode projection shapes (K 7168/4096/3584/2048, N 512..7168) at 1/2/4/8 tokens are bit-identical between the two paths (with FP32 split-K reduction, `B12X_DENSE_SPLITK_TURBO=0`; the bf16 atomic split-K is itself order-dependent). GPU time per projection at 8 tokens (profiler kernel time, i.e. what a CUDA graph pays):

| K x N | separate (us / kernels) | fused (us / kernels) |
|---|---|---|
| 7168 x 1536 | 13.0 / 5 | 10.0 / 2 |
| 7168 x 576 | 9.3 / 5 | 6.4 / 2 |
| 7168 x 896 | 12.6 / 5 | 6.6 / 2 |
| 7168 x 2304 | 10.0 / 5 | 7.2 / 2 |
| 7168 x 3584 | 11.8 / 5 | 10.8 / 2 |
| 7168 x 7168 (stays separate) | 15.7 / 5 | 18.6 / 2 |

`tests/gemm/test_mxfp8_linear_fused_quant_a.py`: bitwise equality eagerly and under CUDA-graph capture across the shapes above and the routing policy (token count, dtype, K padding, N alignment and width, the off switch). The MXFP8/block-FP8 linear suites keep their existing failure set (`test_mm_writes_all_rows_for_unaligned_output_width` at N=132 with more than eight tokens fails before and after; those shapes never take the fused path).

The same change on the Kimi-K3 production snapshot lineage (46c84bab, `b12x/gemm/mxfp8_linear/_kernel.py`) is branch `agent/kimi-k3-mxfp8-linear-fused-quant-a-20260902-served`.

## Summary

- Fuses BF16 activation quantization into `dense_gemm_fused_quant_a` for eligible MXFP8 linear operations.
- Enables the fused path for up to eight tokens, unpadded `K`, and `N` values aligned to 64 tiles up to `B12X_MXFP8_LINEAR_FUSED_QUANT_A_MAX_N` (default: 4096).
- Preserves bit-identical UE8M0 scales and E4M3 values while reducing eligible projections from five kernel launches to two.
- Keeps the existing quantizer and GEMM path for unsupported shapes. Set `B12X_MXFP8_LINEAR_FUSED_QUANT_A_MAX_N=0` to disable the fused path.
- Tests cover eager execution, CUDA-graph capture, routing conditions, and multiple `K`, `N`, and token configurations on SM120/SM121 GPUs.

### DIFF
--- a/b12x/gemm/blockscaled/_linear.py
+++ b/b12x/gemm/blockscaled/_linear.py
@@ -81,15 +81,32 @@ def _source_2d(source: torch.Tensor) -> torch.Tensor:
 _FUSED_QUANT_A_MAX_TOKENS = 8
 
 
+@functools.lru_cache(maxsize=1)
 def _fused_quant_a_max_n() -> int:
     """Widest N routed through the in-CTA activation quantization.
 
     B12X_MXFP8_LINEAR_FUSED_QUANT_A_MAX_N (default 4096): 0 keeps the
-    separate quantizer for every shape. The fused GEMM re-quantizes A in each
-    N tile and has no split-K, so above a few thousand columns the separate
-    quantizer plus split-K GEMM is faster on SM120 (measured at N = 7168).
+    separate quantizer for every shape. Every CTA of the fused GEMM
+    re-quantizes its A rows, so the quantization work grows with the number
+    of output tiles; beyond a few thousand columns that repeated work exceeds
+    the two scale fills and the row quantizer the fused path removes, and the
+    separate path stays faster. The value is process configuration: it is
+    read once and cached (``_fused_quant_a_max_n.cache_clear()`` re-reads
+    it), and it must be a non-negative integer.
     """
-    return int(os.environ.get("B12X_MXFP8_LINEAR_FUSED_QUANT_A_MAX_N", "4096"))
+    raw = os.environ.get("B12X_MXFP8_LINEAR_FUSED_QUANT_A_MAX_N", "4096")
+    try:
+        limit = int(raw)
+    except ValueError as exc:
+        raise ValueError(
+            "B12X_MXFP8_LINEAR_FUSED_QUANT_A_MAX_N must be an integer, "
+            f"got {raw!r}"
+        ) from exc
+    if limit < 0:
+        raise ValueError(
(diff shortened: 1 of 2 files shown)


myshytf:
Review follow-up (cc09c60f):
- applied: `from __future__ import annotations` leads the test module (it was a SyntaxError); `B12X_DENSE_SPLITK_TURBO=0` is assigned unconditionally; the bit-identity test requires a nonzero separate-path result.
- applied: `B12X_MXFP8_LINEAR_FUSED_QUANT_A_MAX_N` is read, validated (non-negative integer) and cached on first use (`_fused_quant_a_max_n.cache_clear()` re-reads it) instead of being parsed per call; the policy test covers the cache and the rejection of bad values.
- applied: the policy docstring explains the N limit by the per-CTA re-quantization cost and no longer claims the fused GEMM has no split-K or quotes an unrecorded measurement.
- 30 passed with `tests/gemm/test_blockscaled.py`.

PR #291, opened 2026-09-02
perf(dense_mla): balance split ranges over the launched CTAs

myshytf:
## Summary

The dense MLA forward kernel assigns every split a fixed run of `chunks_per_split` KV chunks taken from the plan, which is sized for the plan's `max_cache_tokens`. When the caller launches fewer splits than planned, or the live sequence is shorter than the plan, most launched CTAs receive an empty chunk range. On Kimi-K3 decode (TP8/DCP8, 131,072-token local plan: 47 splits × 44 chunks, grid [1, 12, 47]) an 82k-token context is scanned by four CTAs per head tile while 43 sit idle, and `DenseMlaForwardKernel` grows from 0.44 ms to 5.7 ms per decode step over the 24 MLA layers — the whole measured long-context decode slowdown (41 → 46 ms per step at 82k and 163k tokens).

Each launched split now scans `ceil(scan_chunks / active_splits)` chunks of the request's valid range, where `active_splits` is the launched grid and `scan_chunks` the chunks past the window's first valid chunk. The planned `chunks_per_split` only sizes the split scratch. The merge kernel still reduces the launched splits, so the output combines the same partial products through the same log-sum-exp merge and stays within the existing reference tolerance; a launch with one split is unchanged.

## Scope

- `b12x/attention/dense_mla/_forward.py`: balanced per-CTA chunk range; empty splits still publish `-inf` partials.
- No plan, scratch or binding changes. Callers that launch a split-plan prefix for a short live sequence (`_active_dense_mla_splits` in vLLM's `b12x_mla.py`) now need `min(num_splits, live chunks)` splits instead of `ceil(live chunks / chunks_per_split)`; the matching vLLM change is local-inference-lab/vllm `agent/kimi-k3-dense-mla-balanced-splits-20260902-pr`.

## Validation

RTX PRO 6000 (SM120), bf16 KV, 8 heads, page 64, 131,072-token plan (47 splits), all splits launched, forward + merge kernel time per call from the torch profiler:

| live tokens | before | after |
|---|---|---|
| 1,024 | 47.9 us | 7.1 us |
| 4,096 | 126.6 us | 10.5 us |
| 10,240 | 129.3 us | 17.1 us |
| 20,480 | 130.1 us | 26.9 us |
| 40,960 | 134.2 us | 47.9 us |

max |out − reference| is unchanged at every length.

`tests/attention/test_dense_mla.py`: `test_fp8_production_split_plan_handles_short_live_sequence` launches the balanced prefix (five splits for a five-chunk live sequence) and keeps its bit-identity assertion against the full-plan launch; new `test_balanced_splits_cover_a_long_live_sequence_with_every_split` checks a 10,240-token live sequence against the reference with every split and with one split. Run one test per process on this branch (an illegal memory access in the pre-existing `*_tiled_causal_extend_matches_reference` tests poisons the CUDA context of later tests): all of `test_bf16_multi_request_decode_matches_reference`, `test_fp8_production_split_plan_handles_short_live_sequence`, `test_balanced_splits_cover_a_long_live_sequence_with_every_split`, `test_cuda_graph_replay_is_allocation_stable_and_reads_live_inputs`, `test_padded_page_stride_matches_reference`, `test_page_ids_past_int32_scaled_offset_match_reference`, `test_fp8_page_ids_past_int32_scaled_offset_match_reference`, `test_partial_row_budget_changes_native_split_policy` and `tests/attention/test_dense_mla_window.py` (6) pass.

The same change on the served Kimi-K3 kernel lineage (#271's branch, with the sparse chunk selection) is `agent/kimi-k3-dense-mla-balanced-splits-20260902-pr`; its tests pass the same way.

## Summary

- Dense MLA forward now distributes valid cache chunks across active splits using `ceil(scan_chunks / active_splits)`.
- Split ranges account for window offsets and live sequence length. Unused splits publish `-inf` partials.
- Planned cache size still sizes split scratch. Plan, scratch, binding, and merge interfaces remain unchanged.
- One-split launches retain existing behavior.
- Tests cover balanced short and long sequences, all active splits, and one-split equivalence within reassociation tolerance.
- A corresponding vLLM caller update is required for `_active_dense_mla_splits`.
- RTX PRO 6000 measurements improved from 47.9 µs to 7.1 µs at 1,024 live tokens and from 134.2 µs to 47.9 µs at 40,960 live tokens. Reference output differences are unchanged.

### DIFF
--- a/b12x/attention/dense_mla/_forward.py
+++ b/b12x/attention/dense_mla/_forward.py
@@ -75,6 +75,16 @@ class DenseMlaForwardKernel:
         self.qk_dim = int(qk_dim)
         self.value_dim = int(value_dim)
         self.window_size = None if window_size is None else int(window_size)
+        # The windowed producer derives the gathered chunk range from the
+        # tile's first query row (see the range computation in the kernel);
+        # the later rows of a wider tile would see up to query_tile - 1 more
+        # tokens than that range covers. The planner keeps windowed plans at
+        # one row per tile; the kernel refuses anything else.
+        if self.window_size is not None and self.query_tile != 1:
+            raise ValueError(
+                "windowed dense MLA scans one query row per tile, got "
+                f"query_tile={self.query_tile}"
+            )
         self.kv_stages = int(layout.kv_stages)
         self.math_warps = MATH_WARPS_PER_QUERY * self.query_tile
         self.math_threads = self.math_warps * 32
@@ -449,6 +459,8 @@ class DenseMlaForwardKernel:
         first_valid_chunk = Int32(0)
         visible_chunks_end = cache_length
         if cutlass.const_expr(self.window_size is not None):
+            # query_tile == 1 here (enforced at construction), so the tile's
+            # first row is its only row and this range is exact for it.
             tile_local_query = query_start - query_begin
             visible_chunks_end = (
                 cache_length - query_length + tile_local_query + Int32(1)
--- a/b12x/attention/dense_mla/_forward.py
+++ b/b12x/attention/dense_mla/_forward.py
@@ -464,8 +464,24 @@ class DenseMlaForwardKernel:
         valid_chunks = (visible_chunks_end + Int32(CANDIDATES_PER_CHUNK - 1)) // Int32(
             CANDIDATES_PER_CHUNK
         )
-        split_first_chunk = first_valid_chunk + split * Int32(self.chunks_per_split)
-        split_last_chunk = split_first_chunk + Int32(self.chunks_per_split)
+        # Balanced split ranges: the launched splits share this request's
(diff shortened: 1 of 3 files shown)


myshytf:
Review follow-up (17601e54, 68f5c386):
- the gathered range in the windowed path is derived from the tile's first row, but the planner already selects one query row per tile for every windowed plan (`_query_tile` returns 1 whenever `window_size` is set), so wider windowed tiles cannot be planned. The kernel constructor now enforces that invariant (`ValueError` for a windowed kernel with `query_tile != 1`), and `tests/attention/test_dense_mla_window.py` checks both the planner geometry (extend/decode/verify) and the rejection. The unwindowed path gathers the request's full range and masks per row, so it is unaffected (23 passed).

PR #303, opened 2026-09-03
Guarantee KDA prefill pipeline forward progress

voipmonitor:
## Resulting behavior

B12X gated-delta-network attention (KDA) prefill guarantees that preparation
for a tile window completes before the persistent recurrence grid for that
window starts. Preparation for the following window may overlap recurrence for
the preceding window. The two-slot workspace ring is reused only after its
consumer finishes.

The preparation stream uses high priority so preparation for the following
window can make progress while recurrence occupies the device.

## Technical reason

The recurrence kernel polls readiness flags written by the preparation kernel.
Launching the persistent recurrence grid without an explicit same-window
dependency can let recurrence occupy every schedulable cooperative thread array
(CTA), preventing its producer from running. The resulting circular wait can
stall vLLM warmup or CUDA graph capture at full GPU utilization.

CUDA events now encode the producer dependency and the two-window ring reuse
dependency. CUDA graph capture records the same dependencies used by eager
execution.

## Compatibility

The public KDA prefill operation, plan, binding, capacity, and workspace layout
are unchanged. The change affects only stream priority and launch ordering.

## Validation

- Fifteen selected B12X KDA prefill GPU tests passed on an NVIDIA RTX PRO 6000
  Blackwell GPU. A deterministic test uses three populated tile windows to
  exercise ring-slot reuse after window zero is consumed. Eager execution and
  CUDA graph replay match the numerical oracle, preserve recurrent state, keep
  workspace capacity and buffer addresses fixed, and allocate zero bytes during
  replay.
- The vLLM GLM-5.3 serving warmup shape completed with error code zero.
- Server startup and full and piecewise CUDA graph capture completed with tensor
  parallelism of four and decode-context parallelism of one and four.
- Matched cold 32K-token prefill medians at a 4,096-token scheduler budget were
  14,899 tok/s for B12X and 14,900 tok/s for FlashKDA with decode-context
  parallelism of one. With decode-context parallelism of four and full
  cross-rank KV-cache gathering, the medians were 13,106 tok/s for B12X and
  13,242 tok/s for FlashKDA. FlashKDA therefore remains the automatic selection;
  B12X is qualified as an explicit backend.

The [serving qualification receipt](https://github.com/voipmonitor/b12x/blob/6bacce5de2e23dfcec774f38b5aa7a5dec23d25c/docs/evidence/kda_prefill/20260903-glm53-nvfp4-vllm-e2e.json)
records both source trees, clean-worktree state, physical GPU identities, stock
clock mode, exact commands, raw run measurements, correctness results, medians,
and directed comparison ratios.

## Summary

- KDA prefill preparation now runs on a high-priority stream.
- Each recurrence waits for its window preparation to complete before execution.
- Preparation for the next window overlaps recurrence for the current window.
- CUDA events enforce preparation and two-slot workspace reuse dependencies during eager execution and CUDA graph capture.
- Public interfaces and workspace layouts remain unchanged.
- Added three-window eager and CUDA graph coverage for output, recurrent state, fixed workspace capacity, stable buffer addresses, and allocation-free replay.
- Added GLM-5.3 NVFP4 serving evidence comparing B12X with FlashKDA at decode-context parallelism 1 and 4. B12X is qualified as an explicit backend, while FlashKDA remains the automatic selection.

### DIFF
--- a/b12x/sequence/kda_prefill/_cute_kernels.py
+++ b/b12x/sequence/kda_prefill/_cute_kernels.py
@@ -2135,7 +2135,13 @@ def _side_resources(device: torch.device, windows: int) -> _SideResources:
                 "KDA prefill pipeline resources must be created by a warm run before CUDA graph capture"
             )
         if resources is None:
-            resources = _SideResources(stream=torch.cuda.Stream(device=device), events=[])
+            # Give the prepare stream higher priority so preparation for the
+            # following window can make progress while the persistent
+            # recurrence grid consumes the current window. The pipeline below
+            # makes each recurrence depend on its own completed prepare.
+            resources = _SideResources(
+                stream=torch.cuda.Stream(device=device, priority=-1), events=[]
+            )
             _SIDE[device.index] = resources
         current = torch.cuda.current_stream(device)
         while len(resources.events) < needed:
@@ -2150,9 +2156,9 @@ def run_prefill(
 ) -> None:
     """Launch the window pipeline: prologue, then prepare and recurrence per window.
 
-    Prepare launches run on a per-device side stream, recurrence launches on
-    the current stream. A window's recurrence overlaps its own prepare through
-    the ready flags; prepare of window ``w`` waits for the recurrence of window
+    Prepare launches run on a high-priority per-device side stream. Recurrence
+    for window ``w`` starts after prepare ``w`` completes, while prepare
+    ``w + 1`` may overlap recurrence ``w``. Prepare ``w`` waits for recurrence
     ``w - 2`` before reusing that workspace ring slot. Under stream capture the
     fork and join are recorded as graph dependencies.
     """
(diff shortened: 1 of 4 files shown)


voipmonitor:
@coderabbitai review

voipmonitor:
@coderabbitai review

The three-window eager/CUDA-graph ring-reuse test and repository-visible serving receipt are now included. The PR description expands the deployment terms and links the immutable receipt.

voipmonitor:
@coderabbitai review

The requested three-window ring-reuse test, source-locked performance receipt, and context-independent validation description are present at the current head.

lukealonso:
Superseded by `4b3899124`, which guarantees same-window prepare completion before recurrence on `master`. The three-window ring-reuse and CUDA graph regression coverage was retained separately in `9690a83a2`. The priority-stream change and unqualified serving evidence were intentionally not carried forward.

PR #308, opened 2026-09-04
test(kda-prefill): cover near-collinear long-sequence stability

voipmonitor:
Add a 16,384-token KDA prefill stability regression with repeated BF16 keys, a saturated update coefficient, slow decay, and a repeated sixteen-token value block. Exercise the production planned binding and workspace path with both ordinary state slots and real FP32 state element offsets beyond signed 32-bit range.

Poison output, final state, and error status before execution. Require a zero error status and finite, nonzero output and final recurrent state. The test binding helper accepts an explicit state pool so the high-slot case initializes only its live tail without cloning the roughly 8 GiB backing allocation.

This regression checks finiteness and completed writes; numerical accuracy remains covered by the existing oracle comparisons in the same module. Runtime and public APIs are unchanged.

Validation on physical GPU 7, NVIDIA RTX PRO 6000 Blackwell Max-Q (SM120), PyTorch 2.12.0+cu130, CUTLASS DSL 4.6.2:

- `python -m pytest -q -ra -x tests/sequence/test_kda_prefill.py`: 151 passed, no skips or warnings. Includes CPU recurrence/inverse oracles and GPU long-sequence accuracy, graph replay, frozen kernel resolution, and high state-slot addressing.
- Both 16,384-token stability cases passed, each checking 2,097,152 BF16 output elements and 16,384 FP32 final-state elements.
- Assertion controls confirm rejection of skipped writes, zero-only writes, non-finite output, and non-finite state. These controls replace the launcher locally to assess test sensitivity; they are separate from GPU qualification.
- Ruff passed for the changed file.


voipmonitor:
@coderabbitai review

PR #322, opened 2026-09-05
perf(gdn): consume live projection views and bound metadata validation

voipmonitor:
Qwen GDN decode accepts caller-owned packed projection views and strided outputs through the existing plan/bind/run interface. The recurrent update preserves its math and FP32 default state. Plans with at most 32 state-index cells use one metadata-validation CTA; larger plans retain hash-based validation, including transactional invalid-metadata and duplicate-slot checks.

Normalization and metadata strides use a runtime Int64 ABI. Graph capture reuses warmed kernels across changed live capacities, view alignments and strides, including offsets beyond 2**31 elements. Existing contiguous bindings remain valid; no public signatures, defaults or policy schemas change.

Validation: 55 GDN and 7 targeted KDA tests passed on RTX PRO 6000 Blackwell Max-Q, plus Ruff. Coverage includes rollback, wide pool offsets, malformed metadata, frozen compilation, changed-input graph replay, stable addresses and zero replay allocation. The canonical nine-case GDN graph benchmark showed approximately 8–20% lower transaction latency; three noisy cases retained gains on repeat. This measures the GDN component with four planned sequences and four state-index columns; it does not establish full-model serving gains.


voipmonitor:

PR #323, opened 2026-09-05
perf(hyperconnection): read strided projection views without staging

voipmonitor:
Hyperconnection scaled SiLU and residual injection consume row-strided slices of a merged projection directly, allowing callers to remove two staging copies per consumer invocation. Columns remain contiguous, rows must be disjoint, and output alias checks cover the full addressed input span. Existing public signatures and contiguous input behavior are preserved.

CuTe kernels use runtime Int64 row strides and static model-geometry cache keys. Scaled SiLU preserves BF16 rounding before activation; residual combine preserves approximate reciprocal/FMA arithmetic, and combined RMSNorm arithmetic remains unchanged. Warm the kernels before CUDA graph capture.

Validation: 32 tests passed without skips, including the two-device case, exhaustive finite-BF16 SiLU inputs, frozen compilation across row counts/strides, wide input offsets, changed-input graph replay and allocation stability. The staged-versus-direct ABBA component benchmark on RTX PRO 6000 Blackwell Max-Q measured 5.527 to 3.058 us at four rows and 271.152 to 267.056 us at 6019 rows, with bit-exact outputs. These timings include staging and the consumers, not projection GEMMs or full-model serving. Ruff reports one pre-existing SIM103 in an unchanged helper.


voipmonitor:

PR #324, opened 2026-09-05
perf(moe): bind canonical expert scales and support persistent external routes

voipmonitor:
Dynamic NVFP4 MoE bindings consume contiguous FP32 per-expert activation-scale vectors directly from their owning tensors, eliminating two scale refresh copies. Scalar, strided and differently typed inputs retain execution-time normalization into scratch. Binding performs no scale copy, and graph replay observes in-place owner updates.

Grouped NVFP4 external route planning also supports the existing persistent-grid scheduler. Default routing policy, expert IDs, probability/tie semantics, quantization math and shared native A4/A16 weight storage are unchanged. Public signatures and defaults are unchanged.

Validation: 42 binding tests, 10 external-route/scale-update GPU oracle cases across both work schedulers, and one shared-native A4/A16 graph test passed without skips. Coverage includes live mutation, inactive routes, frozen resolution, fixed addresses and allocation-free replay. Ruff and diff checks passed. A public-run graph profile confirmed zero scale copy calls and four GPU launches with canonical vectors, versus two scale copies and six launches with scalar/strided scales. No latency or full-model speedup is claimed.


voipmonitor:

PR #325, opened 2026-09-05
perf(quant): cooperate across lanes for small MXFP8 row batches

voipmonitor:
## Summary

Accelerate standalone MXFP8 activation quantization for linear plans with a row bound of eight or fewer. Eight lanes cooperate on each K32 block in 128-thread CTAs, preserving E4M3 bytes, UE8M0 scales and physical write coverage. Larger planned capacities and native-trellis ordering retain their existing launch configuration.

Dispatch honors expected_m so changing live rows within a planned capacity does not select another compiled callable. Public interfaces and precision policy are unchanged. Small BF16 block-FP8 GEMM plans already fuse activation quantization. Small FP16 block-FP8 GEMM plans use the standalone quantizer and select this path; full FP16 GEMM latency was not measured. This is a standalone quantizer improvement; no BF16-activation serving gain is claimed.

## Validation

46 GPU tests passed on RTX PRO 6000 Blackwell Max-Q with CUTLASS DSL 4.6.2: bit-exact finite BF16/FP16 patterns, quantizer oracles and graph replay, frozen callable reuse across live row counts, and public block-FP8 scratch reuse. No skips; Ruff passed. A regression test exposed and now covers live-row dispatch bypassing expected_m.

Balanced warm CUDA-graph component timings against master ac35c928, BF16 K2560: M1 1.568 -> 0.945 us, M4 3.016 -> 0.949 us, M8 3.135 -> 0.955 us (1.66x–3.28x). Six samples per arm, 512 quantizations per graph, ABBA ordering. Values, both scale layouts and written padding match exactly. M9/M16 controls differ by less than 0.5%. These are standalone quantizer timings, not full GEMM or serving measurements.

## Summary

- Linear MXFP8 inputs with up to 8 planned rows use 8 cooperating lanes and 128 threads for K32 reduction.
- Larger inputs and `trellis_native_mma` inputs retain their existing execution paths.
- The new optional `expected_m` parameter supports planned row capacities without changing default behavior.
- Quantized values, scales, padding writes, graph allocation, and fixed-resolution behavior remain unchanged.
- GPU tests cover finite BF16 and FP16 patterns, odd shapes, larger `k` values, planned-capacity reuse, non-contiguous inputs, and graph replay.
- Results match the scalar reference bit-for-bit.


voipmonitor:

PR #327, opened 2026-09-06
Kimi-K3 prefill: route-block-64 planner fit, stable route pack, dense-GEMM ports

myshytf:
## Behavior

Three prefill-campaign items for the served Kimi-K3 TP9/DCP9 stack (research record: `kimi-k3-production/research/prefill-campaign-20260906/`, reports `moe-kernel.md`, `impl-moe-planner.md`):

- **Route block 64 planner fit** (`42326124`, `ac8572e6`): `compile_w4a16_fused_moe` sized the B stage of its shared-memory fit check at `weight_bits = max(4, trellis_bits)`, rejecting block 64 for the 2-bit trellis (103,424 B estimated vs 101,376 B limit) although the staged layout needs 88,064 B. The fit now uses the layout's real unit bytes (shared with the kernel's `b_unit_bytes`); the residency/grid contract and `_select_tile_config` are unchanged (tests pin both). Block 64 becomes selectable through the existing `VLLM_KQUANT_W4A16_PREFILL_BLOCK_M`. The extent harness validates `--topk-ids` replay metadata (captured routing) and reports the routing histogram.
- **Stable route pack** (`dca22a2d`): atomic scatter plus a per-expert segment sort replaces the whole-workspace scan (896 programs × 73,728 slots, ~153 µs/layer at 4,608 tokens); identical ascending layout, so downstream kernels are bit-identical. `B12X_W4A16_STABLE_ROUTE_PACK_SCAN=1` keeps the scan kernel for A/B.
- **Ports from `master`** (`9b358623` persistent-epilogue TMA-store wait; `6c57554c` + `cd629e54` 188-SM dense plan rules; `6d099ec7`/`cbe8affa`/`c6a4f3da` MXFP8 scale-padding fill removal): applied to the served lineage; the split-K slice rules of the plan ports are intentionally excluded (they change the number of atomic BF16 partials and thus decode numerics). At the served Kimi-K3 prefill shapes 8 of 63 plans change (M=4608 only: o_proj-class → 128×128/BK64, K=7168-class → 128×64), decode/tail plans are unchanged (tests pin both tables). `047f27c5` adds a digest+timing probe of the packed MXFP8 linear at the nine served shapes.

## Reason

Rank-0 traces of the served build show the fused MoE at 88 TFLOP/s (18 % of peak, decode-instruction bound, two trellis decodes per expert per launch at 4,608 tokens) and dense GEMMs at 42-63 % of peak; the route pack is a fixed 153 µs per layer.

## Compatibility

[...]

PR #328, opened 2026-09-06
W4A16 fused MoE: funnel-shift T12 decode chain, cross-tile prefetch, token-major rotation

myshytf:
## Behavior

Three switchable changes to the b12x W4A16 fused MoE prefill kernel used by the served Kimi-K3 TP9 stack (report `kimi-k3-production/research/prefill-campaign-20260906/impl-moe-kernel.md`), all bit-identical by construction and all default on:

- `b27ab012` — the SQG-XOR-Cheb-T12 window decode uses a funnel-shift chain (`bfe`/`lop3`/single `mad.lo` address/`prmt` packing): 17 → 13 PTX and 15.5 → 13.5 SASS instructions per weight, same table bytes read at the same indices. `B12X_SQG_XOR_CHEB_T12_DECODE_CHAIN=legacy` restores the served chain.
- `c81dae5d` — the whole-tile GEMM loop resolves the next tile one step early and issues its first pipeline stages during the current tile's epilogue (the red scratch is separated from the B stages: dynamic smem 72,704 → 86,576 B at block 48, still 1 CTA/SM; block 64 keeps the alias and issues after the drain). Same k-tile order, stage assignment and fold order. `B12X_W4A16_CROSS_TILE_PREFETCH=0` disables it.
- `79da52fd` — in the served configuration (`broadcast_suh=True`, coupled Hadamard) the rotation rows depend only on the token, so the rotation phase writes `M` rows (33 MB) instead of `M·top_k` rows (528 MB) and FC1 gathers `route // top_k`; the row arithmetic is the same `_rotate_coupled_row`, so the MMA operands are identical. `B12X_W4A16_TOKEN_MAJOR_ROTATION=0` disables it.
- `5c58adb2` — a device probe comparing the funnel and legacy decode intrinsics.

With all three switches off the compiled cubin is byte-identical to the served kernel (offline compile of the served cache objects, `evidence/moe-kernel-impl-20260906/offline-compile-summary.json`).

## Reason

The prefill MoE launch is trellis-decode-instruction bound (88 TFLOP/s, 1 CTA/SM); the decode chain, the exposed prologue of every tile and the 528 MB rotation write are the three costs that can be cut without touching the arithmetic.

## Compatibility

Stacked on the served lineage (`feat/kimi-k3-tp9-colocated`, served form of #319); the commits of this PR are the four listed. Switches are compile-cache keys, so both variants coexist in one cache directory.

## Validation

- Offline (serving image, no GPU): CuTe DSL → PTX → `ptxas`/`cuobjdump`: instruction counts and register/smem per variant; CPU tests of the layout regions, tile order, parity bookkeeping and cp.async FIFO model (27 passed), token-major rotation (4 passed); switches-off cubin hash equals the served cubin.
- GPU (maintenance window; `run_gpu_tests.sh prefill-moe-kernel`): extent harness digests of the served tree and of the switch sets legacy/funnel/prefetch/default at m = 1..4608, both widths, must all equal; device tests; `compute-sanitizer` memcheck (m=61, 4608) and racecheck (m=61) clean. Pending — opened for preservation before the window.

## GPU evidence (maintenance window, 2026-09-06 04:24 UTC, one idle RTX PRO 6000; `research/tp9-colocated-qsrt-20260905/evidence/w6-decode_b12x-*`)

- Item 4 funnel-shift decode chain alone: output digests equal to the served kernel at every m (1…4608) and both widths; m = 4,608 graph replay 7,764.9 → 7,361.3 µs (width 384) and 6,104.7 → 5,853.2 µs (256), −5 %.
- Item 5C cross-tile prefetch on top: m = 1…8 passed, m = 16 produced NaN in every element of the graph-replay output (`benchmark_qsrt_tp9_extent.py:146`). Item 5D was not reached. The branch was excluded from the served set as a block (5C and 5D defaulted on).

Root cause (`23a40a55`): an epilogue scratch overrun. Two epilogue stages write from `sh_red_off` with different extents — output staging (`sh_red_size` = 13,056 B) and the CTA-partial fold (16,384 B at the served geometry). The served layout aliases the scratch over the B stages and reserves 16,384 B; the de-aliased layout of the prefetch commit reserved only staging + bias = 13,312 B, so the fold overran 3,072 B into all four scale stages and the first 1,024 B of A stage 0 — exactly the bytes the next tile's `cp.async` stages were already filling. The unordered write-after-write leaves e4m3 scale bytes overwritten by fp32 accumulator words (`0x7F`/`0xFF` = NaN) on tiles where the copy landed first, so every accumulation of the tile is NaN. The geometry does not depend on m; the m ≤ 8 passes were the race not firing. (`assert_close` reports the same message when both tensors are NaN, and the harness records `finite` only after the assertion, so the log does not establish that the eager output was finite.) Fix: the epilogue span is `max(staging + bias, fold)` (+3,072 B per CTA; still 1 CTA/SM under 99 KiB). Offline sm_120a recompilation: with every switch off the cubin is byte-identical to the served lineage base `f0d6081f` (`21f6a056…`); funnel-only `e74226e5…` (the item-4 record), prefetch `e988212e…` (smem 86,576 → 89,648), rotation `0bca50ed…`. A shared-memory writer model driven by the kernel constants shows the overrun (13,312 vs 16,384 → 3,072 B) before the fix and 0 after; 73 CPU tests pass (48 before).

Restructuring (`8f018961`): 5C and 5D default off, so the branch default equals item 4 alone; harness sets `legacy` / `funnel` / `prefetch` / `rotation` measure each switch independently, the sanitizer runs per research set with harness-written markers (the previous `sanitizer_clean` string never appeared in the logs, so that gate could not pass), and the decision rule ships item 4 on its own digest/timing while 5C/5D each need their own set to be digest-identical, not slower than `funnel`, and memcheck/racecheck clean before their defaults change.

## Served (maintenance window 7, 2026-09-06 07:17 UTC, fingerprint `23d121efcdf35e1e`)

Every set is digest-identical to the served kernel at m = 1…4,608 and both widths; m = 4,608 graph replay (width 384 / 256): served 7,762 / 6,094 µs, legacy 7,767 / 6,103, funnel 7,376 / 5,863 (−5.0 %), prefetch 7,246 / 5,769 (−6.6 %), rotation 6,442 / 4,848 (−17.0 % / −20.4 %) (`w7-decode_b12x-moe-kernel-*.json`). Item 4 is served on its evidence. 5C and 5D are served by operator override: their memcheck / racecheck runs did not finish within the window (memcheck at m = 4,608 exceeded 16 minutes and was stopped), so their sanitizer gate is still open and will be rerun at m = 61 and 257; the device tests reported 89 passed / 25 failed, the 25 being pre-existing test-suite setup errors of this lineage (`plan_b12x_fp4_moe_weights() ... 'trellis_codebook'`, `w13_layout 'trellis_t256_proj'`, `weight_layout 'trellis_t256'`) that never reach a launch.

Qualification of the served set (`candidates/k3-tp9-colocated-20260905/evidence/w7/`): cold 8k prefill 2,152 / 2,113 tok/s (1,947 / 1,945 before), cold 64k 2,027 (1,850), c1 decode 95.9 / 74.5 / 74.9 tok/s (96.0 / 79.7 / 71.0; the third prompt at equal acceptance +5.5 %), cross-image isolation, LMCache arms Q1-Q7 and the prefill reproductions all pass. The set combines lab b12x #327, #328, #329 and vllm #677 on the served lineages.

https://claude.ai/code/session_011aPCVZBsteYgs4PTAQYuE5

## Summary

- Adds switchable, default-enabled W4A16 optimizations for SQG-XOR-Cheb-T12 decoding, cross-tile prefetch, token-major coupled-Hadamard rotation, stable route packing, and fused prefill route reduction.
- Adds cache-key coverage for optimization switches and preserves legacy behavior through environment switches.
- Adds route-safety handling for inactive or invalid routes. Invalid routes stage to safe placeholder entries or produce zero contribution without unsafe weight access.
- Replaces BTX trellis preparation with QSRT-based preparation and renames the native layout to `trellis3_t256`.
- Adds Kimi-K3 dense MLA support with fixed 576-wide query/cache records, 512-wide values, sparse chunk selection, query-specific visibility, and token-major cache-stride handling.
- Adds general split-K dense GEMM reduction, FP8 and MX-FP6 preparation paths, fused activation quantization, and improved narrow-N tiling.
- Adds TP9 PCIe collective support, including BF16 two-shot pull/push all-reduce, island9 push all-reduce, TP9 DMA and oneshot routing, graph replay, balanced shard partitioning, and expanded TP8/TP9 support.
- Adds EXL3 K6/MCG small-M GEMM integration and vendored CUDA support for EXL3 dequantization, GEMM, Hadamard transforms, barriers, and memory operations.
- Removes obsolete BTX schema, BTX preparation, strided sparse MLA, static FP8 quantization, and paged sparse-index remapping modules.

## Compatibility

- Existing optimization behavior remains available through environment switches.
- Dense MLA callers must use the Kimi-K3 fixed geometry and updated sparse-attention arguments.
- `PreparedW4A16MoeWeights` and `MoEWeightPreparationPlan` now expose QSRT fields instead of BTX trellis metadata.
- New TP9 runtimes require nine-rank topology support and retain validation for device, shape, dtype, alignment, and launch parameters.

## Validation

- Adds CPU and GPU decode-equivalence tests for legacy and funnel T12 implementations.
- Adds route-safety, fused-sum scratch, token-major layout, cross-tile prefetch, QSRT sharding, and dense MLA regression tests.
- Adds TP9 collective arithmetic, graph replay, physical-system, and BF16 launcher tests with opt-in hardware gates.
- Documents inactive-route sanitizer validation and fused-sum qualification results.
- Provides TP9 collective, QSRT extent, and W4A16 timing/precision benchmark scripts.
- GPU digest, timing, and sanitizer validation for the primary optimization set remains pending where documented.


myshytf:
Lineage note (2026-09-06): a port of this branch onto the lab-facing lineage `feat/kimi-k3-tp9-colocated-on-master` (head of #319, 194 commits over `master` 195e26c5) is pushed as `myshytf:feat/kimi-k3-prefill-decode-chain-20260906-on-master`. It carries item 4 (the funnel-shift decode chain, `b27ab012` ported with the served-only direct-table lines dropped: this lineage has no 64 KiB direct T12 table); with the chain set to `legacy` the offline sm_120a cubin of the fused MoE prefill kernel (width 384, block 48) is byte-identical to the base `ac4587e5` (`21f6a056…`, 289,728 B — the same bytes as the served lineage), and `tests/_lib/test_sqg_t12_decode_chain.py` passes (17). Items 5C and 5D are not ported: they rewrite the whole-tile GEMM loop, which on the served lineage already carries the direct-table staging of #287/#300 that master lacks, so 40 of their 53 hunks do not apply — they follow once #287/#300 are on master.

myshytf:
Publication audit: the prepared master-lineage funnel-chain port is preserved at https://github.com/myshytf/b12x/tree/feat/kimi-k3-prefill-decode-chain-20260906-on-master (c270b9e60649 on ac4587e54543, lab #319 lineage). This branch contains the funnel-chain subset, not a completed master port of every prefetch/rotation switch in the served branch. It is linked here as an alternate-base implementation rather than creating a duplicate PR. The existing served-lineage PR head and its qualification statements remain unchanged.

AI-assisted publication audit at the operator’s explicit request. No code, deployment or merge approval was changed.

https://claude.ai/code/session_01KxvNwugeU8RJFd7WRYwNLG

PR #329, opened 2026-09-06
PCIe DMA ring: static-buffer replay, paired all-gather, column reduce-scatter with fp32 wire

myshytf:
## Behavior

Three additions to the nine-rank PCIe DMA ring runtime (`b12x/comm/pcie/pcie_dma.py`, `_dma_kernels.py`) for the served Kimi-K3 TP9 prefill (report `kimi-k3-production/research/prefill-campaign-20260906/impl-collectives.md`, design `collectives.md` §7):

- `9340d3af` — graph replay over one static buffer per all-reduce shape: replay entries keyed `(op, numel, dtype)` with a borrowed static output, an op-sequence eviction guard (`REPLAY_EVICTION_GUARD_OPS`), and `all_reduce_input` / `borrow_output` / `is_ring_storage` so the producing GEMM writes into the ring's static input and the consumer reads the static output (removes the staging copies: 18 ms per 4,608-token chunk).
- `91cca004` — `all_gather_pair()`: two rank-local blocks (Kimi-K3 router fp32 logits 17 MB and routed latent 33 MB per layer) gathered in one copy-engine ring pass on the ring stream, so shared experts and top-k on the main stream run underneath; bit-identical (copies).
- `3a1f0fd8` — `reduce_scatter_columns(inp, wire, cols)`: `[rows, K]` bf16 reduced as `[9, rows, cols]` column blocks over eight ring steps; `fp32` wire keeps fp32 partials on every hop after the first (one bf16 rounding instead of eight; link bytes 1.67 N vs 1.78 N for the all-reduce), `bf16` wire is the served precision class. New CuTe mixed-precision add launches; flag-slot ranges AR 0..144 / AG pair 152..161 / RS 168..186 with a per-world-size disjointness assert. `benchmarks/precision_pcie_tp9_allreduce.py` gains `dma-ring`, `dma-rs-bf16`, `dma-rs-fp32`, `--kernels`, `--norm` (RMSNorm variance against the fp64 reference).
- `08833185` — nine-GPU qualification test (`tests/comm/test_pcie_dma_ring_ops_tp9_gpu.py`): in-place replay + borrow (300 replays, counter monotonicity), gather-pair eager + replay against `torch.distributed.all_gather`, reduce-scatter fp32/bf16 blocks against a chained reference with fp64 mismatch counts, and an AR→AG→RS→AR 200-layer sequence replay; stages are written to a JSON evidence file.

## Reason

Per 4,608-token prefill chunk on rank 0, 56 % of the critical path is collectives; the ring already runs at 99 % of its bandwidth bound, so the remaining exact levers are fewer bytes (reduce-scatter where the consumer is row-parallel), copy-engine gathers overlapped with compute, and no staging copies.

## Compatibility

Stacked on the served lineage (`feat/kimi-k3-tp9-colocated`, served form of #319); the commits of this PR are the four listed. Every new path is opt-in from the caller (vLLM side: lab vllm PR on `feat/kimi-k3-prefill-20260906`, env-gated, default off); with the gates off the ring behaves as before.

## Validation

- CPU (serving image): a nine-thread host emulation of the ring (`tests/comm/pcie_dma_emulation.py`: real flag protocol and add arithmetic over host memory) — replay/borrow consistency, gather-pair byte equality, both reduce-scatter wires bit-equal to the chained reference, fp32 mismatch < ring/4, slot disjointness — all passed (`evidence/collectives-impl/b12x-cpu-tests.log`).
- GPU (maintenance window; `run_gpu_tests.sh physical`, `prefill-dma-ring-ops`, `prefill-latent-precision`): stages above on nine GPUs; precision harness at rows 4608 × width 3584 (`--norm`) and 7168: `dma-rs-fp32` must show fewer mismatching elements, no larger max ulp and lower rel-L2 than `dma-ring`, with zero cross-rank inconsistency, before the fp32 wire is enabled in production. Pending — opened for preservation before the window.

## GPU evidence (maintenance window, 2026-09-06 04:25 UTC, nine idle RTX PRO 6000; `research/tp9-colocated-qsrt-20260905/evidence/w6-comm_b12x-*`)

- Item 7 (`item7_in_place_replay_and_borrow_passed`, 302 replays, 33 slots) and item 1 (`item1_gather_pair_eager_and_replay_passed`) passed in two runs of `tests/comm/test_pcie_dma_ring_ops_tp9_gpu.py`.
- Item 3 failed in both runs on the first replays (`fp32 wire block differs (call 1)`; the other run recorded a bf16-wire fault at call 2), and the abort stopped the test before the layer-sequence stage. The precision step aborted because `PCIeTwoShotBF16.all_reduce` rejects rows 4608 × width 3584.

Root cause (`afd270da`): a missing stream-ordering edge in the reduce-scatter schedule. The running sums alternate between two partial buffers (`partials[step % 2]`); step k's send reads `partial_piece(k-1, p)` on the copy stream while step k+1's add writes the same buffer on the main stream, and nothing ordered them (the copy stream waits on the add, the flag wait binds a rank only to its predecessor's copies). A rank whose outgoing hop is slower than the chain feeding it overwrites a payload it is still transmitting; eager launches are throttled by the host (the CuTe launcher costs tens to hundreds of µs per kernel, comparable to a 3.69 MB piece transfer), a replayed graph issues at device speed — hence eager passes, replays fail, and the corrupted block is another rank's result (the "rank 8" in the traceback is the first error `mp.spawn` dequeued; the SIGTERM list shows other ranks exited on their own). The all-reduce is immune (its 2(world−1)-step schedule rewrites a chunk eight steps after the send that read it); the paired all-gather forwards into a different scratch area per step. Fix: before the add of step k ≥ 2 that writes a partial, the main stream waits on the already-recorded copy-engine event of step k−1 for the same piece. No arithmetic, slab layout, slot range or replay-arena change.

CPU proof (`03915f60`): the emulation gained a happens-before model (vector clocks on streams/events, the flag protocol's cross-rank edges, graph replay brackets, per-kernel byte-range access declarations; `EmulatedRing(model=True).conflicts`). Nine ranks, cols 400, two pieces, eager + capture/replay + replay: reduce-scatter fp32 and bf16 wires 486 unordered write-after-read pairs before the fix → 0 after; layer sequence 243 → 0; all-reduce and paired all-gather 0 both before and after (calibration); block values bit-identical to the chained reference. `tests/comm`: 504 passed / 31 skipped / 2 pre-existing failures (world-size-9 expectations in `test_pcie_dma_kernels.py`).

Harness (`a919b7f0` + campaign files): the ops test records a differing block as a fault with every rank running the same number of ring ops, runs the layer sequence without the reduce-scatter if it failed, and writes the evidence before raising; the precision harness skips a kernel that rejects the size with a recorded reason; `prefill-latent-precision` runs `dma-ring,dma-rs-bf16,dma-rs-fp32` at both widths. Next window: `prefill-dma-ring-ops` (four stages, `ops = ar, ag_pair, rs_fp32`) and `prefill-latent-precision` (w3584 `--norm`, w7168).

## Served (maintenance window 7, 2026-09-06 07:17 UTC, fingerprint `23d121efcdf35e1e`)

All four nine-GPU stages of `tests/comm/test_pcie_dma_ring_ops_tp9_gpu.py` passed after `afd270da`: in-place replay + borrow (302 replays, 33 slots), gather pair, reduce-scatter — fp32 wire **2 mismatching elements against the fp64 sum out of 16,515,072** (served ring 7,486,238; bf16 wire 7,500,268) — and the 200-layer sequence with ops ar / ag_pair / rs_fp32 / ar (`w7-comm_b12x-prefill-dma-ring-ops.log`). Items 7, 1 and 3 (fp32 wire) are served with their env gates on. The precision harness step did not produce a record: its acceptance wrapper reported "the PCIe DMA ring could not be built" for every dma kernel and swallowed the exception (harness defect, to be surfaced); the item-3 decision used the ops-test fp64 comparison.

Qualification of the served set (`candidates/k3-tp9-colocated-20260905/evidence/w7/`): cold 8k prefill 2,152 / 2,113 tok/s (1,947 / 1,945 before), cold 64k 2,027 (1,850), c1 decode 95.9 / 74.5 / 74.9 tok/s (96.0 / 79.7 / 71.0; the third prompt at equal acceptance +5.5 %), cross-image isolation, LMCache arms Q1-Q7 and the prefill reproductions all pass. The set combines lab b12x #327, #328, #329 and vllm #677 on the served lineages.

https://claude.ai/code/session_011aPCVZBsteYgs4PTAQYuE5

## Summary

- Adds opt-in TP9 PCIe collectives with CUDA Graph replay, static buffers, paired all-gather, and column reduce-scatter.
- Supports BF16 and FP32 reduce-scatter wires with mixed-precision kernels and ordering protection.
- Adds TP9 support across one-shot, two-shot BF16, island, DMA, and DCP A2A runtimes.
- Adds CPU emulation with happens-before conflict detection and tests for replay, ordering, precision, flag counters, and composed sequences.
- Existing all-reduce behavior remains unchanged unless callers select new paths.

## Attention, GEMM, and MoE

- Reworks dense MLA for fixed Kimi-K3 geometry, sparse chunk selection, per-query cache lengths, and FP32 split partials.
- Adds narrow-N FP8 split-K policies, fused MXFP8 activation quantization, and the EXL3 Trellis K6/MCG CUDA extension.
- Replaces BTX checkpoint handling with QSRT atom preparation and TP9 sharding.
- Adds `trellis3_t256`, EXL3, QSRT E4M3, and SQG FP16 D3L support.
- Adds stable route packing, W4A16 prefill fused-sum reduction, and inactive-route validation.
- Removes obsolete static FP8, BTX, and legacy sparse trellis paths.

## Validation

- TP9 GPU qualification covers collectives, replay, precision, and physical-device sequences.
- Precision benchmarks skip unsupported kernels collectively and preserve evidence from later kernels.
- BF16-wire replay mismatches are recorded without stopping subsequent FP32 and sequence checks.
- GPU qualification and precision runs remain environment-gated.


myshytf:
Lineage note (2026-09-06): this branch stacks on the graph-replay runtime of #281 (`agent/kimi-k3-dma-ring-graph-replay-20260902-pr`), which is not on `master` or on `feat/kimi-k3-tp9-colocated-on-master`; a port onto the lab-facing lineage needs #281 merged or stacked first (the first commit's three conflict hunks are exactly the replay runtime). The served-lineage branch here is the review vehicle until then.

PR #334, opened 2026-09-06
W4A16 fused MoE: pipeline-depth and small-M CTAs-per-SM launch knobs (opt-in)

myshytf:
## Behavior

Two opt-in launch knobs of the W4A16 fused MoE kernel (`b12x/moe/_shared/kernels/w4a16/kernel.py`), both defaulting to the served geometry:

- `B12X_W4A16_STAGES` (default 4, 2..8): depth of the cp.async A/B/scale pipeline of every W4A16 launch, read once at import so the shared-memory footprint functions and the kernel's constexpr pipe loops agree. K-tile `i` keeps pipe `i % stages` and the fold order, so the depth changes no arithmetic order; it trades shared memory for latency hiding.
- `B12X_W4A16_SMALL_M_BLOCKS_PER_SM` (default 1, 1..4): raises the one-CTA-per-SM pin of the block-8 (decode) launch up to the register and shared-memory limits. The small-M work loop is grid-agnostic, so the grid size changes no arithmetic order either.

## Reason

On the Kimi-K3 TP9 decode step (rank-0 trace, `kimi-k3-production/research/prefill-campaign-20260906/decode-anatomy-20260907.md`) the fused MoE launch is 40 % of the step at ~120 µs per layer with one 8-warp CTA per SM, latency-bound. The 64 KiB direct T12 table only fits next to a pipeline footprint of ≤ 35,840 B; the block-8 footprint is 39,056 B at 4 stages, 30,352 at 3, 21,648 at 2 (56,464 at 6, 73,872 at 8; offline sm_120a compiles), so the depth also decides whether the direct table is staged. The knobs let an idle-GPU measurement rank {stages × CTAs per SM × table} on the real extent before any of it reaches production.

## Compatibility

Defaults reproduce the served geometry (4 stages, pin 1); no kernel code path changes without the environment. The cubin of the decode geometry is byte-identical for pin 1 and pin 2 (the pin only changes the persistent grid).

## Validation

- CPU (serving image): `tests/moe/test_w4a16_decode_launch_knobs.py` — 13 tests (parsing, bounds, the module constant, the pin and its resource limit) pass.
- Offline compiles of the decode geometry at 2/3/4 stages and at two CTAs per SM succeed with the footprints above.
- GPU: pending — harness step `decode-extent-variants` (`research/tp9-colocated-qsrt-20260905/run_gpu_tests.sh`) measures ranks 4 and 0 at m = 1/2/4/8 for variants served/s2/s3/s6/s8/b2/notable/notable-b2/notable-s8-b2; a variant ships only when its digests equal the served launch's at every m and width and it is ≤ 0.98 × the served time at m = 4 on every rank, and the prefill launch is not slower than 1.02 × served in the integration check.

https://claude.ai/code/session_015BejFtqwvuJKpxWH8cuHCh


myshytf:
Nine-GPU measurement (2026-09-06 16:08 UTC, rank 4's 256-width extent, layer 1, block 8, m = 1/2/4/8, 200 graph replays): every variant reproduces the served launch's output digests at every m and width; graph-replay time at m = 4 relative to the served configuration (4 stages, one CTA per SM, direct 64 KiB T12 table): 3 stages 1.010, 2 stages 1.157, 6 stages 1.618, 8 stages 1.642, two CTAs per SM 1.001, direct table off 1.603 (with two CTAs per SM 1.603; with 8 stages 1.643). The served configuration is the fastest of the set — the direct table is worth 37 % of the decode launch and deeper pipelines evict it — so the knobs stay opt-in measurement aids. Records: kimi-k3-production/research/tp9-colocated-qsrt-20260905/evidence/w9-decode-decode-extent-r4-*.json.

PR #339, opened 2026-09-07
Optimize Kimi decode collectives and audit QSRT verify kernels

myshytf:
Kimi-K3 TP9/DCP9 decode spends time gathering paired projections, selecting experts, and running separate quantization or output-cast kernels. This branch provides opt-in transport and kernel choices for those operations, plus numerical and measurement tooling. The branch uses the serving source lineage rooted at `71df6033`; full integration into another lineage requires matching its collective and binding contracts.

GitHub reports conflicts with `master`. The published evidence applies to the serving-lineage source identified here; a merge-resolution port requires its own compatibility checks.

## Behavior

- Query head gather accepts aligned, non-overlapping padded batch rows while retaining compact pull/push staging. `supports_all_gather_heads_output(out)` lets integrations avoid an intermediate only for supported layouts; compile ABI 14 carries an Int64 output stride. Padding is preserved.
- `B12X_SQG_XOR_CHEB_T12_DIRECT_PAIRS=1` returns four independent byte pairs from the direct shared LUT. The native full-rotation path remains FP16 internally; model-facing BF16 input and the separate BF16 fragment conversion remain supported. The option is off by default.
- `B12X_W4A16_PHASE_PROFILE=1` records per-CTA phase timestamps and FC1 reduction-turn polling in a preplanned workspace tail for isolated single-tier decode. Compiled launch capacity checks and pool-mode invalidation protect the workspace contract. It is off by default and requires GPU qualification.

- Paired projection gathers can use posted-write push transport. `B12X_PCIE_DCP_A2A_PAIR_TRANSPORT` selects pull or push independently; unset inherits the general DCP transport. Logical output widths skip trailing rank padding without changing valid columns.
- The fused paired-gather/router path can keep top-16 selection in registers. Scores, tie order, route weights, and their sequential FP32 normalization retain the scan path's contract.
- Shape-selective fused MXFP8 activation quantization supports decode projections beyond the default output-width cap. Each shape needs a measured comparison before deployment.
- `B12X_W4A16_TOPK_SUM_OUTPUT=bf16` or `fp16` rounds the full-rotation route sum once in the store. Compile keys, scratch, output buffers, and binding checks follow that dtype; unset retains FP32 output.
- `B12X_W4A16_M8_CTA_THREADS` selects 256 or 512 threads for the supported MoE geometry. The 512-thread option remains performance-unqualified in the DFlash2 composition.
- Collective tests cover graph replay, clipped outputs, and worker failure propagation. The QSRT extent benchmark records raw replay samples, operating state, FP32 diagnostic errors, and optional four-row launch attestation. `--server-metrics-url` marks service overlap and leaves missing observations explicitly unknown.

## Offline verify-kernel qualification

Source `1414b128` was compiled with CUDA 13.3.73, Torch 2.13.0 and CUTLASS DSL 4.6.2, without GPU device mapping. The four baseline CUDA `.text` sections match the archived production-contract objects byte for byte. Exact hashes are in `validation/performance/kimi_k3_verify_direct_pairs_offline_20260908.json`.

| Width | CTA | SASS instructions, baseline → pairs | Registers/thread, baseline → pairs |
| --- | --- | --- | --- |
| 256 | 256 | 6104 → 5976 | 120 → 121 |
| 384 | 256 | 6128 → 6000 | 121 → 119 |
| 256 | 512 | 4728 → 4664 | 100 → 103 |
| 384 | 512 | 4768 → 4704 | 98 → 103 |

Lookup, FP8 conversion and HMMA counts stay constant; no local loads/stores appear. Three register increases remain flagged for device testing. Static instruction reductions are not latency measurements.

Validation of the added paths: 102 CPU collective tests and 60 targeted CPU MoE tests pass. The CPU-only run skipped nine GPU decoder tests; all nine subsequently passed on an RTX PRO 6000 Max-Q. Both nine-rank pull and push collective gates passed, including graph replay, raw 656-byte query records and output rows beyond 2 GiB. The summaries are retained in `validation/performance/kimi_k3_verify_gpu_correctness_20260908.json`. All eight MoE width/CTA/pair combinations, an instrumented case, four TP9 gather specializations (pull/push × eager/graph), and six 2/3/4-bit × FP16/BF16 oracle specializations compile offline. The decoder oracles have executed successfully; the fused native-MoE variant and phase-record checks remain pending. Ruff still reports the base's undefined legacy helper in `intrinsics.py` and two `zip(strict=...)` findings in `_impl.py`; changed standalone files and tests pass.

Native-MoE equality, phase-record GPU validation, full-model evaluation and default decode-benchmark measurements for these changes remain pending. `benchmarks/benchmark_qsrt_verify_variants.py` preserves an interleaved native-extent harness with shared weights, mutated inputs and exact-object evidence; it is research-only and has not yet run on a device. See `docs/dcp_query_output_layout.md` and `docs/qsrt_verify_diagnostics.md` for contracts and qualification commands. The gather integration is in [vLLM #654](https://github.com/local-inference-lab/vllm/pull/654). The direct LUT consumer extends #328's modal-table decoder, and the stride ABI extends #335's transport; their existing work is retained in this PR rather than duplicated.

## Recorded device qualification (2026-09-07)

The paired-gather and routing checks recorded on nine idle RTX PRO 6000 GPUs on 2026-09-07 include 90 CPU collective tests, TP9 pull/push and captured-graph checks, and four GPU top-16 equivalence cases. Four-row paired-gather replay changes from 19.0 to 10.1 microseconds, and top-16 selection from 6.21 to 3.36 microseconds under that geometry. These operation-level results do not establish a universal model speedup. The fused-quantization shapes and output-store routes have separate compile, dtype, and binding tests; stale baseline binding tests are documented in the existing validation discussion.

The actual QSRT K2 source extents at logical TP9 ranks 0 and 4, widths 384 and 256, are preserved in [the extent qualification document](https://github.com/myshytf/b12x/blob/9807c62e/docs/evidence/kimi_qsrt_extent_diagnostics.md) and its JSON record. All 16 cases across M=1/2/4/8 and 256/512 threads are finite, eager/graph identical, and have matching cross-thread output digests. Four-row traces attest the actual launch widths, registers, grid, and shared memory. The record identifies the frozen source pair and the limits of single-extent validation.

**There is no qualified 512-thread speed comparison.** Three of four replay intervals show live request activity. The rank-0 512-thread interval has idle endpoint gauges and unchanged counters, but its 256-thread reference overlaps serving. Raw timings remain diagnostic and are excluded from performance decisions. The routing/overlap-classification CPU suite passed **eight tests**, with Ruff and diff checks passing for that source.

The extent benchmark allocates output in the configured dtype, as implemented in revision `ab269462fcfc`; instrumentation retains that contract. Packed MLA evidence belongs to [B12X #311](https://github.com/local-inference-lab/b12x/pull/311), and combined DFlash2 serving qualification belongs to [vLLM #704](https://github.com/local-inference-lab/vllm/pull/704). The invalid 128 Ki benchmark baseline is excluded; its request-ownership correction is [llm-inference-bench #16](https://github.com/local-inference-lab/llm-inference-bench/pull/16).

AI assistance was used for implementation, measurement, and evidence review, including GPT-5.6 Terra at maximum reasoning effort. Maintainer review is required before merge.


myshytf:
## Maintenance window 12 (2026-09-07 15:12–16:27 UTC): C1 + C2 served

**Decision**: ship C1 (paired projection gather push transport) + C2 (register-held top-16 selection); the logical-width paired gather kernel (`d0a78754`) is included but inactive until the vLLM side passes widths (local-inference-lab/vllm#702). Candidate fingerprint `f9ceef2e2c829901` (b12x `b699401e` on the served vLLM `f76b70768a`, environment of the window-11 decision), accepted into the production cache namespace `iso-a14668fc43a57358` (byte copies and a selection-order change only; identical expert ids, weights and latent bits proven on the idle-GPU harness).

**Idle-GPU proofs (nine RTX PRO 6000)**
- `dcp-a2a-push`: TP9 physical test and the generic nine-rank test passed under pull and push, including the new paired-gather checks (batches 1..64, clipped rows, eager, adjacency and CUDA-graph replay).
- Paired gather graph replay at the served decode geometry (µs): batch 1 11.1 → 6.7, batch 4 19.0 → 10.1 (×0.53), batch 8 28.3 → 15.4.
- `decode-router-topk`: 4 passed (scan and register selections agree bit for bit on normal, tied, non-finite, constant and signed-zero keys at 128/256/512 threads); kernel 6.21 → 3.37 µs.

**Serving qualification** (all passed): canaries, concurrency-1 samples, cold 8k/64k TTFT probes, cross-image vision isolation, LMCache arms Q1–Q7, prefix-hit/external/mixed prefill reproductions, prefill tail integrity, 220k cold prefill with a clean kernel log.

**Benchmark pair** (`llm_decode_bench`, concurrency 1, same host, before = served build `d0f0b8355e85576c` measured in the same window; validity: all contexts cc=1, no queue, no competing request):

| context | decode steps/s before → after | prefill tok/s before → after |
|---|---|---|
| 0 | 30.1 → 30.9 (×1.026) | – |
| 8k | 27.5 → 28.1 (×1.023) | 1957 → 1947 (×0.995) |
| 16k | 25.6 → 26.1 (×1.022) | 2026 → 2019 (×0.997) |
| 32k | 25.5 → 26.1 (×1.023) | 2056 → 2050 (×0.997) |
| 64k | 25.4 → 26.0 (×1.022) | 2011 → 2005 (×0.997) |
| 128k | 24.8 → 25.3 (×1.021) | 1913 → 1907 (×0.997) |

Decode steps/s +2.1…2.6 % at every context (token/s follows the draft acceptance of each 30 s sample and is not the metric); prefill within 0.5 % (the kernels are decode-only).

**Test-harness fixes on this branch found by the window**: the nine-rank GPU test's slot-parity probe raced against peer pushes (`2ef1ba81`), worker exceptions now fail loudly instead of hanging the group (`a7194e88`), and the paired checks run last (`f223c54f`, `c18ed4e3`). The MoE measurement steps (`benchmark_qsrt_tp9_extent.py --widths`, `b699401e`/`bcc42717`) run in the next window.

Served since 2026-09-07 16:11 UTC.

myshytf:
## Maintenance window 13 (2026-09-07 16:40–17:08 UTC): G2 served, G4a closed, K-lever data

**Idle-GPU proofs (one RTX PRO 6000)**
- `decode-topk-sum-output`: 37 passed — the bf16 and fp16 in-kernel stores of every full-rotation top-k sum variant (plain H128, coupled Hadamard with per-expert and broadcast output scale, mapped/unmapped, int32/int64 ids, m = 1/4/9) equal the separate cast byte for byte → `B12X_W4A16_TOPK_SUM_OUTPUT=bf16` shipped (removes the per-layer cast kernel, 93 per decode step).
- `decode-mxfp8-linear-ab` (G4a, 48 cells): digests equal for every shape/m/L2 state (both routes bit-identical), but at the served m = 4, L2-hot, the fused route is slower for the KDA in-projection (×1.23), KDA out-projection (×1.20), MLA out-projection (×1.00) and the layer-0 MLP (×1.02–1.38); it wins only at m = 1 (×0.76–0.93) and for the side-stream shared-expert down projection at m = 4 (×0.89). Serving list left empty; the env stays available.
- `decode-moe-baseline`: isolated graph replay of the served fused MoE decode launch (layer-1 extents, served switches): 384-wide 46.4 / 65.0 / 114.2 / 212.2 µs at m = 1/2/4/8 (Zipf routing within 3 %), 256-wide 39.5 / 52.5 / 87.2 / 151.8 µs; graph = eager bitwise; ≈1.5 µs per expert block over a ≈22 µs floor.
- `decode-moe-ncu` (m = 4, 384-wide): issue slots 22 %, IPC 0.91/SM, 77 % of cycles with no eligible warp at 2 active warps per scheduler, 8.7 cycles per issued instruction, L1/TEX (shared-memory lookup) pipe 62.7 %, DRAM 22.6 %, L2 11 %, 121 registers, one CTA per SM (smem). The kernel is latency-bound at eight warps per SM; the shared-memory pipe leaves ≈1.6× headroom — the basis of K1 (`04f246d7`, 512-thread M8 CTA, same K-slices and fold order; ptxas 96/92 registers, no spills), proven and measured in the next window.

**Served** since 2026-09-07 16:52 UTC: fingerprint `ccebff1df704ecc9` = b12x `8cf3bc4d` (window-12 kernels + `B12X_W4A16_TOPK_SUM_OUTPUT=bf16`) with vLLM `c7c17309b4` (local-inference-lab/vllm#702: logical-width paired gather rows, bf16 route-sum buffer). Qualification passed (canaries, cold TTFT probes, cross-image vision, LMCache arms Q1–Q7, prefill reproductions, tail integrity, 220k cold prefill with a clean kernel log).

**Benchmark pair** (before = the window-12 build measured 40 min earlier on the same host; valid matrix):

| context | decode steps/s before → after | prefill tok/s before → after |
|---|---|---|
| 0 | 30.9 → 31.4 (×1.018) | – |
| 8k | 28.1 → 28.5 (×1.016) | 1947 → 1954 |
| 16k | 26.1 → 26.5 (×1.015) | 2019 → 2017 |
| 32k | 26.1 → 26.5 (×1.014) | 2050 → 2053 |
| 64k | 26.0 → 26.4 (×1.014) | 2005 → 2008 |
| 128k | 25.3 → 25.6 (×1.013) | 1907 → 1910 |

Cumulative since the campaign's served baseline (`d0f0b835`): decode steps/s 30.1 → 31.4 at ctx 0 (×1.043) and 24.8 → 25.6 at 128k (×1.032), prefill unchanged.

myshytf:
## Maintenance window 14 (2026-09-07 17:29–17:52 UTC): K1 proof not run; served set rebuilt on this branch's head

The window followed a host reboot (GPU 0 switched to compute mode with ECC off). Its idle-GPU steps failed immediately for two harness reasons, both fixed on this branch: the extent benchmark allocated its route-sum output as fp32 while the container environment carried `B12X_W4A16_TOPK_SUM_OUTPUT=bf16` (`ab269462`: the benchmark now follows the setting; the bind-path error names the expected dtype), and the quantizer test step mounted a source worktree without the compiled `vllm._C*.so` (harness now uses the served export). The decision rule booted **without** K1 (env unset, arithmetic identical to window 13): fingerprint `69bd7b59d4fe9968` = window-13 set on b12x `04f246d7` + vLLM `c7c17309b4`, tail-trace env removed.

Reduced qualification at the operator's request (canaries, prefill tail integrity, kernel log since boot — the other arms were not run). Benchmark vs the window-13 build: decode steps/s ×0.983…0.985 at every context (31.4 → 30.9 at ctx 0), prefill ×0.99–1.00. The shortfall is a clock policy artifact of the reboot, not the kernels: the static clock policy left GPU 4 locked at its 3,090 MHz table maximum (the other eight ranks ran unlocked with a 3,337 MHz ceiling); under load GPU 4 held 2,340 MHz against 2,520–2,587 MHz on the others and 2,580 MHz before the reboot, and TP9 paces at the slowest rank. The lock was reset (`nvidia-smi -i 4 -rgc`); the K1 proof (control vs `B12X_W4A16_M8_CTA_THREADS=512` digests and replay times, stall-reason ncu) and a clean benchmark pair run in the next idle window.

myshytf:
K1 (`B12X_W4A16_M8_CTA_THREADS=512`) and the direct-LUT pair decode measured on an idle GPU (maintenance window W-1, 2026-09-09 07:5x KST; tree `perf/k3-moe-decode-arms-20260909` = verify-kernels `2e98180e` + `ab269462` + the interleaved harness `ffa5ee30` + two harness repairs; GPU 5 alone, P1, SM 2,400–2,407 MHz, memory 16,365 MHz, no throttle reason, layer-1 real QSRT weights, served switches incl. the 64 KiB direct table):

`benchmark_qsrt_verify_variants.py`, m = 4 graph replay, 200 interleaved samples per arm (Latin-square order); all four arms byte-identical (eager, graph, mutated inputs) at m = 1/2/4/8:

| extent | cta256 (served) | cta256 + pairs | cta512 | cta512 + pairs |
|---|---:|---:|---:|---:|
| rank 0, 384-wide, uniform | 122.50 µs | 121.95 (0.996) | 126.82 (1.035) | 125.86 (1.027) |
| rank 0, 384-wide, Zipf | 121.94 | 121.82 (0.999) | 126.32 (1.036) | 125.86 (1.032) |
| rank 4, 256-wide, uniform | 93.25 | 93.22 (1.000) | 95.36 (1.023) | 95.26 (1.022) |
| rank 4, 256-wide, Zipf | 93.22 | 93.09 (0.999) | 95.23 (1.022) | 95.87 (1.028) |

Nsight Compute (m = 4, 384-wide, profiler clock 2.17 GHz): 256 threads 138.1 µs, 512 threads 139.1 µs; executed instructions 48.9 M vs 56.9 M (+16 %); registers 121 vs 98, no spills; stall per issue-active: MIO throttle **3.44 vs 7.08**, barrier 1.17 vs 2.66, long scoreboard 0.66 vs 0.75, short scoreboard 0.51 vs 0.89; L1/TEX pipe 62.5 % vs 64.9 %, DRAM 23 %.

Reading: the decode launch is bound by the shared-memory instruction pipe of the direct-table lookups (MIO throttle), not by latency; doubling the warps lengthens that queue and duplicates per-warp A-fragment/metadata work. **K1 is held** (slower at both extents) and the pair decode is neutral (within noise) — neither enters the serving environment. The remaining exact lever for this launch is fewer table lookups per weight, not geometry.

Evidence: `kimi-k3-production/research/upstream-research-20260909/w1/evidence/w1c/` (decision.json, the four variants JSONs with GPU state and per-sample timings, `w1b-k*-decode-moe-ncu*.csv`). Harness repairs on the measurement branch (`a6c8bbc8`, `1095f13a`): the DSL `dump_to_object` export is optional evidence (it fails to re-parse cached objects under cutlass 4.6.2), and arms are identified by their compile record because this lineage does not attach `fused_launch` to the binding.

PR #343, opened 2026-09-08
PCIe collectives: trigger programmatic dependents at kernel entry (Kimi-K3 TP9 decode)

myshytf:
## Behavior

The six PCIe collective kernels of the Kimi-K3 serving lineage (one-shot, fused one-shot + RMSNorm, FP8-wire two-shot, BF16 two-shot scatter / pull / push) execute `griddepcontrol.launch_dependents` as their first statement. A kernel launched behind them with the programmatic-stream-serialization attribute may start while the collective waits on its peers; its own `griddepcontrol.wait` still orders its reads of the collective's output after the grid completes. Kernels launched without the attribute serialize as before. There is no switch: the trigger is inert without attributed dependents. Arithmetic, barriers and buffer contracts of the collectives are unchanged.

Branch: `perf/k3-twoshot-pdl-trigger-20260909` = served lineage `0bf9f177` + one commit (`4cde442a`). Same mechanism as the one-shot early trigger of #341 and the opt-in release of #279, applied to the kernels the TP9 decode path uses (two-shot push at the decode payloads, one-shot below 16 KB).

## Technical reason

On the served Kimi-K3 TP9/DCP9 decode step the three all-reduces of every layer cost 4.7 ms of a 30.8 ms verify graph (rank-0 trace, 64 Ki context; `kimi-k3-production/research/verify-critical-path-20260908`), most of it barrier waits with idle SMs. The vLLM consumers that follow them — `_attn_res` and the KDA metadata kernels — are already launched with the attribute and call `gdc_wait` before their first dependent read (audit of every `launch_pdl=True` site in vLLM `f9ac0209a2`), so the trigger lets their prologues overlap the fabric round trips. #279 measured +4.46 % verifier steps/s for the same mechanism at GLM-5.3 TP4.

## Compatibility

No configuration change; b12x kernels are not launched with the attribute themselves and keep serializing. Requires a CuTe DSL with `cute.arch.griddepcontrol_launch_dependents` (the serving image's cutlass 4.6.2 has it).

## Validation

Status: **implemented; CPU-validated; GPU ordering test and serving measurement pending an idle nine-GPU window** (`kimi-k3-production/research/upstream-research-20260909/w1/`).

- `tests/comm/test_pcie_collectives_pdl_trigger.py`: the trigger is the first statement of all six kernels (source check, CPU).
- `tests/comm/test_pcie_pdl_dependent_tp9_gpu.py` (+ `tests/comm/pdl_dependent.py`): every collective runs with a waiting dependent copy kernel behind it, attributed and not, eagerly and inside CUDA graphs on nine GPUs; both outputs are compared bitwise with the run alone. Not yet run (needs nine idle GPUs).
- `tests/comm` on the CPU of the serving image: 531 passed, 35 skipped, 2 failed (`test_pcie_dma_kernels.py` world-size review cases, identical failures on the unchanged served tree).
- No performance claim: the served-step effect is measured only by the cc = 1 decode benchmark after a candidate boot.

https://claude.ai/code/session_01DCXE5V6K1RjMqcw89ijGBb


myshytf:
GPU qualification (maintenance window W-1, 2026-09-09 07:4x KST, nine idle RTX PRO 6000, served target stopped; tree `4cde442a` on the served lineage `0bf9f177`):

- `tests/comm/test_pcie_pdl_dependent_tp9_gpu.py`: **passed** on nine GPUs — two-shot pull, two-shot push and one-shot all-reduces each ran with a waiting dependent copy kernel behind them, launched with and without the programmatic-stream-serialization attribute, eagerly and inside CUDA graphs (40 iterations per shape); the collective output and the dependent's copy equalled the run alone bitwise in every case.
- `tests/comm/test_pcie_tp9_physical.py` with this tree: every stage passed (one-shot, two-shot pull/push balanced partition, island9 push, DMA wire tail, DCP gather/LSE).
- `benchmarks/benchmark_pcie_tp9_allreduce.py`, this tree vs the served kernel tree (the six one-shot/two-shot source files are identical between the served lineage and the control tree apart from this change): graph-mode ratio over 55 b12x cells median **0.996** (max 1.08 at two-shot 1 row × 4,096, min 0.867), eager-mode median 1.004. The trigger costs nothing measurable in isolation.
- No serving claim yet: the step-level effect (the PDL-attributed vLLM consumers overlapping the fabric wait) is measured only by the cc = 1 decode benchmark pair after a candidate boot.

Evidence: `kimi-k3-production/research/upstream-research-20260909/w1/evidence/w1c/` (decision.json, harness logs `w1b-pdl-*`, `w1b-ctl-allreduce-bench.json`).

myshytf:
Serving measurement (maintenance window W-2, 2026-09-09 08:12–08:4x KST): candidate `d2b74a24a779125c` = the served DFlash2 UVA-writes configuration with this tree (`4cde442a`) as the only change, isolated cache namespace.

- Qualification passed (cold 11,918-token request, 9,216-token external cache hit, two-turn conversation, document canary; finite log probabilities; mean accepted length 2.63).
- Operator benchmark, cc = 1, 30 s per context: steps/s 30.9 / 29.9 / 29.3 / 28.4 / 29.1 / 27.3 at 0 / 8k / 16k / 32k / 64k / 128k; prefill 1,929 / 1,997 / 2,030 / 1,984 / 1,885 tok/s.
- Same-boot baseline (served set, four contexts measured before the boot): steps/s ratio **1.005 / 1.005 / 1.006 / 0.979** at 0 / 8k / 16k / 32k — no measurable gain (the 32k baseline cell had an anomalous accepted length of 2.04). The pre-registered rule (>= 0.99 at every context) was not met; the served set was restored.
- Reading: the vLLM kernels that follow these collectives with the programmatic-stream-serialization attribute (`_attn_res`, the KDA metadata kernels) execute `gdc_wait` as their first statement, so an early trigger gives them nothing to overlap. The gain reported for the same trigger in #341 / #279 came from the weight-first consumer that stages its weights before the wait. This PR therefore stays a producer-side prerequisite; a serving gain needs consumers with pre-wait work (MXFP8 dense GEMM weight/descriptor staging, the fused MoE table staging, the router GEMV).

Evidence: `kimi-k3-production/research/upstream-research-20260909/w2/evidence/w2/` and `candidates/k3-pdl-trigger-20260909/`.

PR #350, opened 2026-09-09
PCIe DMA ring: row-granule chunk mapping and fp32 reduce-scatter hops on the served ring (port of #331)

myshytf:
## Summary

Ports the opt-in row-count-invariant chunk mapping (`B12X_PCIE_RING_GRANULE_ROWS`) and the fp32 trailing reduce-scatter hops (`B12X_PCIE_RING_FP32_HOPS`) of #331 onto the ring as served by the Kimi-K3 TP9/DCP9 deployment (`0bf9f177`: static replay entries keyed by op tag, in-place all-reduce entry, nine-rank wire padding, paired all-gather and column reduce-scatter sharing the slab). #331 is based on the lineage before that restructuring and does not apply textually (eight conflicting regions in `pcie_dma.py`), so this is a port, not a rebase; the mapping, the hop arithmetic, the reference module and the step-table proofs are #331's unchanged.

Why: a prefill chunk run as two row halves reproduces the whole chunk's bits only if every all-reduce sums each element in the same order regardless of the row count. The served mapping's order depends on `row // (rows / world)`; the granule mapping makes it depend on `row mod (world * g)` only. The fp32 hops make the deployed set measurably more precise (#331).

## What changes

- `_all_reduce_aligned` dispatches to `_all_reduce_lossless` when the tensor takes the granule mapping or the channel has fp32 hops. Buffer ranges come from `pcie_dma_reference.lossless_access_plan`; `B12X_PCIE_RING_CHECK_BOUNDS=1` validates them against the slab before any kernel is issued.
- Replay key `("ar", numel, dtype, granule_elems)`: equal-sized tensors of different widths get separate entries, and the in-place entry captures the lossless schedule.
- The scratch slab grows by one shard capacity per fp32 hop; the lossless path addresses it through `_scratch_offsets`, every other op keeps its linear `_scratch_ptr` areas, which the layout always covers (ops on one channel never overlap).
- The nine-rank padded wire (1-D) keeps the served mapping with the channel's fp32 hops.
- The mixed-precision adds are the column reduce-scatter's; `prepare_reduce_scatter(wire="fp32")` runs at construction when hops are on.
- `tests/comm/pcie_dma_emulation.py` builds rings with `granule_rows` / `fp32_hops`.
- Defaults (`g = 0`, `p = 0`) issue the served kernel sequence unchanged.

## Validation (serving image, CPU, no CUDA)

- `tests/comm/test_pcie_ring_granule_cpu.py`: 60 passed (#331's step-table proofs).
- `tests/comm/test_pcie_ring_granule_replay_cpu.py` (new, 13 cases): the served ring code over the host-memory emulation: reference agreement for granules 1/2/4 x fp32 hops 0/1/3/7, row halves bit-identical to the whole (and not with the served mapping), the in-place replay entry reproduces eager and is keyed by granule, padded wire with fp32 hops, the schedule model reports no unordered overlapping access (fp32 stage, widened receive areas, in-place static buffer), the bounds check accepts every issued plan.
- `tests/comm/test_pcie_dma_replay_cpu.py`: 43 passed (key literals carry the granule element).
- `tests/comm` in full: 620 passed, 35 skipped, 2 failed: `test_pcie_dma_kernels.py::test_dma_rejects_unreviewed_world_sizes_before_backend_setup[9]` and `::test_dma_reviewed_world_sizes_match_tp10_accumulator_limit`, which fail identically on `0bf9f177`.
- Nine-GPU proof pending an idle window: `benchmarks/benchmark_pcie_ring_granule_tp9.py` (carried over). #331's window-7 run ended in an `unspecified launch failure` on rank 8 after every variant's graph was captured; the bisect plan is one variant per process with `B12X_PCIE_RING_CHECK_BOUNDS=1`, then `compute-sanitizer` on the smallest failing configuration.

AI-assisted (Claude Code); the operator reviews before merge.

https://claude.ai/code/session_016V6akAFmEqoNQhkkAca6x7


myshytf:
Nine-GPU results (maintenance window W-3, 2026-09-09, evidence `research/prefill-2x-20260909/w3/evidence/w3-20260909/` in the production repository):

**Harness defects found first.** `benchmarks/benchmark_pcie_ring_granule_tp9.py` rebound each variant's graph input/output tensors to two loop-local names; `torch.cuda.graph()` empties the caching allocator on entry, so the earlier variants' buffers were unmapped and their warm-up replay faulted (`illegal memory access` / `unspecified launch failure`) on whichever rank reached it first, for every hop configuration, the granule-only one included. This is the failure of #331's window-7 run. Fixed in `34cbe6c6` (buffers held per variant, `graph_matches_eager` recorded); `f1b0f7a3` fixes the hop-count parsing that raised for `dma-ring` once the timing loop completed; `93bc977a` / `18a78c17` make `test_pcie_dma_ring_ops_tp9_gpu.py` compare against the channel's own mapping and key.

**Ring benchmark, granule 128 (median µs, rel-L2 vs fp64, split-mismatch elements):**

| shape | served dma-ring | granule 128 | granule 128 + 1 fp32 hop |
|---|---|---|---|
| 4608×7168 | 4310 · 3.702e-3 · 15,336,401 | 4292 · 3.702e-3 · **0** | 4556 · 3.344e-3 · 0 |
| 2304×7168 | 2155 · 3.702e-3 · 7,663,883 | 2165 · 3.701e-3 · **0** | 2299 · 3.344e-3 · 0 |
| 4608×3584 | 2154 · 3.702e-3 · 7,663,883 | 2178 · 3.699e-3 · **0** | 2313 · 3.340e-3 · 0 |
| 2304×3584 | 1101 · 3.701e-3 · 3,831,547 | 1109 · 3.697e-3 · **0** | 1175 · 3.339e-3 · 0 |

Every replayed graph reproduces its eager call bit for bit. Granule 256 (also measured) is bit-identical only at 4,608 rows and makes the 2,304-row ring a one-piece ring (+13 % / +22 %), so 128 is the configuration carried forward. The full fp32 wire (7 hops) halves the error at +43 % time; one hop costs +6 %.

**Served ring-ops test** under `B12X_PCIE_RING_GRANULE_ROWS=128`, `B12X_PCIE_RING_FP32_HOPS=1`, `B12X_PCIE_RING_CHECK_BOUNDS=1`: all stages pass (in-place replay and borrowed outputs, paired all-gather, column reduce-scatter on both wires with 0 replay faults, 200-layer AR/AG/RS sequence replay).

Next in the same window: the split-prefill candidate boot with residual digests (vLLM #719).

myshytf:
Nine-GPU proofs on the served rig (window W-3, 2026-09-09; evidence under `kimi-k3-production/research/prefill-2x-20260909/w3/evidence/w3-20260909/`):

- Granule 128 (`B12X_PCIE_RING_GRANULE_ROWS=128`): the ring reduces the 2,304-row halves of a 4,608-row all-reduce bit-identically to the whole on every tested shape (`split_mismatch_elements` 0), with time within ±1 % of the served mapping. Granule 256 also reduces exactly but is 13–22 % slower on 2,304-row halves (a one-piece ring), so 128 is the configuration carried forward.
- One fp32 trailing hop (`B12X_PCIE_RING_FP32_HOPS=1`): relative L2 error of the reduction −9.7 % against an fp64 reference at +6 % time.
- The served ring-ops test passes under both settings; the unsplit production numerics under granule 128 + one fp32 hop pass the serving qualification, and the Kimi-K3 split prefill built on this mapping is bit-identical to the unsplit chunk in every forward (vLLM #719).

The mapping is carried by the split-prefill candidate, which is research-only for now (exact but not faster than the served set), so the defaults stay unchanged here.

PR #353, opened 2026-09-10
perf(moe): split-materialized NVFP4 prefill (1.31x geo on GLM-5.3 shape, 2.21x on large-M receipt)

L14nY1Wang:
## Summary

Adds a **split-materialized NVFP4 MoE prefill path** to the dynamic backend. The monolithic `MoEDynamicKernelBackend` (a single cooperative route/pack + FC1 + FC2 kernel) is split into:

1. a **route/pack-only** cooperative front-end (routing + input quantization + task publication), then
2. two **standalone, non-cooperative, fixed-grid** CuTe-DSL compute kernels — `Nvfp4MaterializedPhase1Kernel` (FC1 + SiLU + NVFP4 requant) and `Nvfp4MaterializedPhase2Kernel` (FC2 + top-k weighted BF16 atomic scatter) — using the SM120 native NVFP4 block-scaled QMMA (`kind::mxf4nvf4 block_scale scale_vec::4X m16n8k64`).

Splitting the GEMM bodies out of the cooperative kernel removes the register/shared-memory union that caps the monolithic kernel, raising SM occupancy on large-M prefill. A repository-visible, reproducible receipt for the split's target band ships with this PR at `docs/evidence/nvfp4_split/`.

## Why (design)

- **phys-major contract, front-end unchanged.** The split phase kernels consume the *existing* expert-major packed-A + F8_128x4-swizzled SFA domain exactly as the monolithic route/pack publishes it (each routed token is quantized once and fanned out to its expert physical rows); `token_map` is used only for FC2 scatter. **No store-side change** to the shared front-end.
- **Original (non-repacked) weights.** The split consumes the original swizzled FP4 payloads + compact F8_128x4 SFB planes; alpha (FC1 dequant), a2 global scale (intermediate requant), and down_alpha (FC2 dequant) are folded identically to the monolithic path.
- **Graph-safe, Int64, frozen resolution.** All three launches live inside one `@cute.jit __call__` on one stream (a single-op CUDA graph captures all three). Pool-scaled offsets are Int64; live task counts are runtime Int32 scalars driving dynamic grids/masks (never in the compile key).

## Correctness gates (before any perf claim)

- Chain vs `moe_reference_nvfp4` oracle: **cos > 0.9999**, abs bound derived from the BF16 output floor the pipeline actually shares with the monolithic `sC` staging.
- **split == monolithic under real per-expert `alpha` / `down_alpha` / `global_scale`** (catches operand-fold/ordering errors a gs=1 domain can't).
- CUDA-graph capture of cooperative+phase1+phase2 → **bitwise-identical replay**, zero replay-time allocation.
- **Live row past the Int32 offset boundary.** `test_nvfp4_phase_live_row_beyond_int32_offset` parks the live domain at a high `tile_base` so the kernels' own scaled offsets (`physical_row * words_per_row` for the payload, `sf_atom * k4 * 512` for the scale plane) exceed `2^31` while only the pool tail is written, then re-checks the reference metrics. Shifting merely the base pointer of a small pool does not exercise the kernels' offset math and is not what this test does.
- Frozen kernel resolution: one compiled callable reused across two live task counts with fixed capacity, proven to materialize strictly more work on the larger count.

### Committed receipt (reproducible)

`docs/evidence/nvfp4_split/` (JSON + generated README) records the command, `argv`, commit, worktree, active/idle GPU mode, package versions, source SHA-256, raw per-iteration samples, and the correctness outcome for both arms. Both arms drive the production `@cute.jit` host adapter `_DynamicMoELaunch`; they differ only by `materialize_intermediate`.

- Hardware: **RTX 5090**, active mode **P1 / ~2925 MHz / 13.8 Gbps** (clocks not locked).
- Config: K=4096, I_tp ∈ {1024, 2048}, E ∈ {8, 64}, top_k ∈ {2, 8}, M ∈ {2048, 4096, 8192}; 100 iterations × 5 interleaved rounds, medians.
- Gate: finite + nonzero outputs, cos > 0.9999 vs oracle, split-vs-monolithic cos > 0.9999, RMSE ≤ `max(8e-4, 3·2⁻⁸·max|oracle|)`.

| E | K | I_tp | top_k | M | split (µs) | monolithic (µs) | speedup |
|---:|---:|---:|---:|---:|---:|---:|---:|
| 8 | 4096 | 2048 | 2 | 2048 | 8837.3 | 20023.4 | 2.27x |
| 8 | 4096 | 2048 | 2 | 4096 | 17129.5 | 38447.4 | 2.24x |
| 8 | 4096 | 2048 | 2 | 8192 | 33800.7 | 75134.8 | 2.22x |
| 64 | 4096 | 1024 | 8 | 2048 | 19270.0 | 42651.7 | 2.21x |
| 64 | 4096 | 1024 | 8 | 4096 | 36359.0 | 78447.1 | 2.16x |
| 64 | 4096 | 1024 | 8 | 8192 | 70757.6 | 151509.2 | 2.14x |

**Geomean 2.207x** over 6/6 qualified engaged shapes.

### Original GLM-5.3 shape sweep (kept for the wider mix)

Hardware: **RTX 5090** (SM120, 170 SMs), single GPU. Shape: GLM-5.3 Flash (shape-only synthetic experts), K=4096, I_tp=2048, E=288, top_k=8, `nvfp4`, `silu`, fast-math, `--reference none`, 5 repeats × 20 iterations, L2 flush 192 MiB. Both arms run back-to-back in one session. `ratio` below is **split/monolithic time** (lower is better).

CUDA graph (`--cuda-graph --graph-mode single-op`):

| batch | monolithic (µs) | split (µs) | ratio (split/mono) | engaged? | split speedup |
|---:|---:|---:|---:|---|---:|
| 512   | 2658.6 | 2651.1 | 1.00 | no (small tile → monolithic) | 1.00x |
| 2048  | 3144.2 | 3143.7 | 1.00 | no (M64 tile, gated out) | 1.00x |
| 8192  | 10079.3 | 7165.4 | 0.711 | yes | 1.41x |
| 16384 | 19370.5 | 14923.8 | 0.770 | yes | 1.30x |
| 24576 | 28489.8 | 21858.3 | 0.767 | yes | 1.30x |
| 32768 | 36823.5 | 29807.6 | 0.809 | yes | 1.24x |

No CUDA graph (`--no-cuda-graph`, matches the Phase-0 baseline harness):

| batch | monolithic (µs) | split (µs) | ratio (split/mono) |
|---:|---:|---:|---:|
| 8192  | 10083.2 | 7171.1 | 0.711 |
| 16384 | 19376.0 | 14934.0 | 0.771 |
| 24576 | 28494.9 | 21858.1 | 0.767 |
| 32768 | 36822.3 | 29915.0 | 0.812 |

- Engaged large-M (M128) shapes: **geomean 1.31x speedup (23.6% less time)**, identical under graph and no-graph (0.764 vs 0.765 time ratio).
- Peak single point **1.41x** (bs=8192); **1.24x** at bs=32768.
- The split engages only when the tile planner selects (128,128) (large prefill); smaller shapes keep the monolithic path.

### Reconciliation

The two measurements do not conflict. The committed receipt isolates the **large-M M128 band** (E=8/64, top_k=2/8) and reports 2.14–2.27x; the GLM-5.3 sweep averages a **wider engaged-shape mix** (E=288, top_k=8, batch 512–32768, CUDA-graph capture), where many physical tiles carry fewer rows and the split's advantage is smaller, giving 1.31x. Use the committed receipt's per-shape, correctness-gated ratios for the covered large-M band; the GLM-5.3 figures describe the wider mix.

## Validation

- Full local regression: **62 passed / 1 skipped** (63 collected) —
  `test_nvfp4_phase_kernels` (5) + `test_nvfp4_split_backend` (6) + `test_nvfp4_split_dispatch_policy` (13) = 24;
  `test_w4a8_dynamic_kernel` (15) + `test_dynamic_w4a8_trellis` (11) = 26;
  `test_nvfp4_auto` (12) + `test_moe_eager_prefill` (1) = 13.
- Code-review passes applied across three review rounds: gated the unvalidated M64 multi-tile specialization out of the predicate (it produces NaN on multi-tile experts — those shapes fall back to monolithic); fixed plan/launch tile-decision parity; removed dead params/fields/imports; corrected a misleading SFA lane-mapping comment (in intrinsic + both kernels); cached the dispatch env flag at import so the launch path never re-reads it; strengthened the receipt's correctness gate (finite/nonzero, split-vs-monolithic agreement, timings withheld on failure) and replaced the base-shift Int64 test with the live-row boundary test.

## Notes / caveats

- **Default**: `B12X_NVFP4_DYNAMIC_MATERIALIZED` auto-enables for matching shapes (mirrors the W4A8 split). Set it to `0` to force the monolithic NVFP4 path for troubleshooting. The flag is **read once at import and cached**, so it must be set before the process starts; it is not re-evaluated per launch.
- **Policy layer untouched**: no `MoeDecodeConfig` schema bump and no embedded-profile change (the split is selected by a structural predicate + env default evaluated in the `_impl` dispatch, so it also works on GPUs without an embedded profile, e.g. this RTX 5090 running in HEURISTIC mode).
- **Pre-existing, unrelated failure**: `tests/policy/test_component_catalog.py::test_every_planned_op_has_exactly_one_policy_registration` still fails (missing `sequence.gdn_prefill` / `sequence.kda_prefill` registration) — not touched by this PR.
- Timings are on **shape-only synthetic experts** (no checkpoint on this box) and `--reference none` (flashinfer JIT is unavailable here); correctness is gated against `moe_reference_nvfp4`, not flashinfer. Re-measure on the target checkpoint before quoting a serving number.

## Files

12 changed, +10476 / −121:

- `b12x/_lib/intrinsics.py` — SM120 native NVFP4 QMMA intrinsic
- `b12x/moe/_shared/kernels/nvfp4_phase1.py` (new) — FC1 + SiLU + NVFP4 requant
- `b12x/moe/_shared/kernels/nvfp4_phase2.py` (new) — FC2 + top-k weighted BF16 scatter
- `b12x/moe/_shared/kernels/dynamic.py` — split specialization wiring
- `b12x/moe/fused_moe/_impl.py` — dispatch predicate, workspace plan, launch parity
- `tests/moe/test_nvfp4_phase_kernels.py` (new)
- `tests/moe/test_nvfp4_split_backend.py` (new)
- `tests/moe/test_nvfp4_split_dispatch_policy.py` (new)
- `benchmarks/benchmark_nvfp4_split_materialized.py` (new) — split vs monolithic A/B receipt
- `benchmarks/gen_nvfp4_split_readme.py` (new) — receipt → README renderer
- `docs/evidence/nvfp4_split/README.md` (new) — evidence receipt
- `docs/evidence/nvfp4_split/20260910-rtx5090-nvfp4-split-prefill.json` (new) — raw samples + hashes

> Note: the `+10476` is dominated by the raw receipt JSON (~6.6k lines); the code+tests portion is ~3.8k.

## Summary

Adds a split-materialized NVFP4 MoE prefill path for validated large-prefill M128 configurations.

- Adds SM120 native NVFP4 QMMA support.
- Adds cooperative route/pack input processing.
- Adds phase 1 for FC1, SiLU, and NVFP4 intermediate materialization.
- Adds phase 2 for FC2 and top-k-weighted BF16 output scatter.
- Preserves packed-A and SFA contracts.
- Supports CUDA graphs, Int64 offsets, frozen kernel resolution, and caller-owned workspaces.
- Uses the monolithic path for unsupported shapes.
- Disables with `B12X_NVFP4_DYNAMIC_MATERIALIZED=0`.

The path improves qualified RTX 5090 cases by 1.24–1.41×, with a committed large-M receipt reporting a 2.207× geomean. Independent serving validation reported approximately 8.6–8.9% higher GLM input throughput and a 9.57% Qwen TP1 increase.

Validation passed 62 tests, including all 24 NVFP4 split tests. Tests cover graph replay, allocation ownership, Int64-boundary addressing, routing, quantization, workspace behavior, dispatch policy, and frozen resolution. Model-shaped numerical checks met the configured cosine and normalized-RMSE bounds. Strict split-versus-monolithic cosine of 0.9999 was not met for some large GLM and Qwen shapes, so the change does not claim bit-exact logits or general model equivalence.


voipmonitor:
Independent serving validation on RTX PRO 6000 Blackwell Workstation, 600 W limits, VRAM **+6000** (not stock). All comparisons use the same physical GPUs per model.

GLM checkpoints contain exactly uniform expert input scales **as vectors**. #354 plus vLLM #727 expose a preparation-time immutable-scale proof, without replacing those vectors or changing checkpoint values; that makes this PR's split path eligible. Your commits through `9a03a9a9` are retained with original authorship in the [complete tested B12X integration](https://github.com/voipmonitor/b12x/tree/release/jovian-nvfp4-split-r33-20260910). This PR remains the kernel review unit; #354 does not duplicate the kernels.

Packaged TP4/DCP1 GLM, FP8 KV, budget 4096, OMP1, 16 NCCL channels/2 MiB, full-and-piecewise graphs. Cold 32K input, one excluded warmup +30 seconds, zero cached tokens:

| Mode | R32 input tok/s | Split image input tok/s | Change |
|---|---:|---:|---:|
| No-spec | 15,737 | 17,128 | +8.84% |
| MTP3 | 15,276 | 16,637 | +8.91% |
| DFlash2 K7 | 15,569 | 16,910 | +8.61% |

All three modes pass three long literal-document lookups with processed logprobs. **39 installed CPU/GPU contracts pass**, including graph replay, caller-owned allocations and Qwen TP1's width-640 geometry. The separate Qwen source-overlay TP1/MTP3 comparison is 15,707 → 17,211 input tok/s (+9.57%); budget 6019. Qwen TP2 width 320 is ineligible and no split gain is claimed there.

Decode is not a general speedup claim. MTP3 verifier C1/C8 changes are +0.99%/−0.32%; output varies with acceptance. DFlash C1 has 96.65–101.02 steps/s across R32 controls versus 97.27–97.32 in the candidate; the lower state reproduces on R32. All observations are retained.

Important numerical limit: independent model-shaped oracles pass the established cosine ≥0.999 / normalized-RMSE ≤0.03 bounds, with smaller split error. Strict split-versus-monolithic cosine 0.9999 does not pass on the large GLM/Qwen shapes. No bit-exact logits or broad model-quality/distribution equivalence is claimed. [Committed commands, exact revisions, raw warm/cold samples, GPU UUIDs and numerical limits](https://github.com/voipmonitor/b12x/blob/perf/nvfp4-immutable-input-scales/docs/nvfp4-immutable-input-serving.md).

L14nY1Wang:
@voipmonitor Thanks for running the independent PRO 6000 validation and for the recognition of the work — much appreciated.
I'll work through the review comments as quickly as I can so the PR is ready to merge into mainline.
Happy to pick up anything this surfaces on the kernel side — e.g. additional shapes or a re-measure against the target checkpoint — if that would be useful for the record.

lukealonso:
Pushed the update at `027cea46d50ae8173ad28a770c7e244e8b9ce0ff`.

- Split FC1 now preserves the configured SiLU clamp.
- Dynamic down-scaling and forced swapped operands retain the compatible monolithic path rather than selecting an incompatible split kernel.
- vLLM allocation ownership is unchanged: b12x plans report scratch requirements and bindings map caller-provided storage. No runtime pool-growth path or blanket intermediate reservation was added. The earlier internal workspace-pool transition repro is not the vLLM caller-owned contract.
- Added real public-path regressions for clamping, option compatibility, and allocation-free plan/bind/run/capture-body/replay under frozen kernel resolution across 512 -> 256 -> 512 live rows.
- Replaced the forced four-cluster benchmark with production planning and CUDA graph replay. The covered production-sized cases selected 188 active clusters per arm.

Validation: **158 passed, 1 skipped** (Qwen checkpoint unavailable); changed Python files pass Ruff. The kernel/source fixes are in `66b0c675`; the subsequent commits add explicit GPU-condition qualification and its receipt.

The [replacement receipt and overview](https://github.com/L14nY1Wang/b12x/blob/027cea46d50ae8173ad28a770c7e244e8b9ce0ff/docs/evidence/nvfp4_split/README.md) identify clean source commit `27977c440d5ce8acd04a3e57efd51384e99c9c4b`; all 67 source hashes matched independently. These are **Max-Q diagnostics with fast math disabled**, not formal release or serving evidence: masks are restricted to 0x0/0x4, initial-to-active transitions are explicit, active arms must remain P1 on the same GPU/memory clock, and the declared SM-clock delta bound is 5%. Five shapes qualify, with a **1.330x geometric mean (monolithic time / split time)**. The sixth shape exceeded the clock bound (7.0647%); its headline timing is withheld and raw samples remain available. The older 2.207x restricted-grid receipt is retained for audit, not credited as production-plan evidence.

voipmonitor:
Independent validation of D-Rock's SwiGLU report on RTX PRO 6000 Workstation:

| Runtime | Clamped NVFP4 oracle |
|---|---|
| Published community R34, unchanged | FAIL: cosine 0.929721, RMSE 0.365815, max error 3.113281 |
| Same R34 with only the 13-line/two-file clamp correction | PASS |
| Public JJ/PR composition, unchanged image runtime | PASS |

R34 uses the same affected B12X revision as R33. The test exercises production
fused-MoE dispatch with a tight limit of 2; the GLM checkpoint config uses 10.
Finite-output checks alone do not catch the missing clamp.

The merged correction in this PR is already in the B12X base for the composition
tracked by [vLLM #731](https://github.com/local-inference-lab/vllm/issues/731).
All four split-dispatch tests also pass, including graph/scratch reuse. We keep
the split implementation enabled; no replacement fix PR is needed.

[Immutable identities, minimal A/B results and raw receipts](https://github.com/local-inference-lab/blackwell-llm-docker/blob/9275b68ca51e79218bb28f38f37ca69eb9a0f6a3/recipes/glm53/r34-preservation.md).
This proves the kernel defect and correction, not that every reported production
runaway has this single cause. We did not repeat D-Rock's production-shaped soak
or attribute the originating error to LMCache.

PR #354, opened 2026-09-10
perf(moe): share NVFP4 inputs for immutable uniform expert scales

voipmonitor:
## Behavior

Allow one NVFP4 input quantization to serve all routed experts when a prepared
weight owner explicitly guarantees immutable input scales and its expert scale
vector contains exactly equal, finite, positive values. Preserve the canonical
vector and direct bindings; do not average scales or requantize checkpoint weights.

Preparation performs the equality check once. Binding reads metadata only.
Versioned mutations invalidate the proof; inference tensors require the caller's
immutability guarantee for the complete owner and CUDA-graph lifetime. Mutable
callers retain per-expert packing. The default API behavior is unchanged.

The dynamic and compact-micro custom-op schemas now mark only the actual
scratch/output writes. Declaring every argument mutable incremented read-only
scale versions and invalidated this proof after execution.

## Relationship to split prefill

This PR is independently based on B12X master. It does not copy or replace
L14nY1Wang's [split-prefill kernels in #353](https://github.com/local-inference-lab/b12x/pull/353).
That PR requires shared-input eligibility; GLM and Qwen checkpoints store equal
scales as vectors, so they did not qualify through the scalar-only check.
The [vLLM companion #727](https://github.com/local-inference-lab/vllm/pull/727)
declares immutability for loader-owned inference weights.
Merge the three review units together for the measured serving optimization.

## Validation

[Serving commands, exact source boundaries, GPU UUIDs and raw warm/cold samples](https://github.com/voipmonitor/b12x/blob/perf/nvfp4-immutable-input-scales/docs/nvfp4-immutable-input-serving.md)
are committed with the PR. They include every observed GLM C1 timing state and
the strict cross-kernel cosine limitation. The complete two-layer integration
also passes 39 installed CPU/GPU contracts, including Qwen width-640 graph replay.

Implemented; qualified locally on RTX PRO 6000 Workstation with VRAM +6000.
`python -m pytest -q tests/moe/test_nvfp4_shared_input_scales.py` passes **15 tests**
against this master-based branch in the R32 dependency image. Changed-file Ruff
and `git diff --check` pass. Tests cover unequal/invalid scales, one-ULP mismatch,
mutable and inference tensors, canonical storage, hot-path value-read exclusion,
mutation invalidation, and actual PyTorch custom-op version accounting.

Serving A/B against the preserved R32 composition, with #353 and the vLLM
companion applied: cold 32K GLM TP4/no-spec prefill **15,842 → 17,152 tok/s
(+8.27%)**, with a second reference at 15,737; Qwen TP1/MTP3 **15,707 →
17,211 tok/s (+9.57%)**. Same physical devices within each comparison. These
are client timings, not isolated-kernel ratios. No decode speedup is claimed.
GLM uses a 4,096-token scheduler budget; Qwen uses its qualified 6,019-token budget.

Independent numerical-oracle and graph replay checks pass. Split versus
monolithic cosine is approximately 0.99989 for these shapes, below a diagnostic
0.9999 cross-kernel threshold; split has lower error against the independent
oracle. This is **not bit-exact parity or a claim of unchanged target logits**.
Literal document/cache checks pass; broad task-quality equivalence is not established.
GLM speculative release guards are reported separately with the image evidence.

DeepSeek's MXFP4/MXFP8 shared-input path already exists; this PR does not claim a
DeepSeek speedup. Qwen TP2's 320-wide expert partition does not qualify for #353.

AI-assisted implementation and validation, submitted for human maintainer review.


voipmonitor:
vLLM adapter: https://github.com/local-inference-lab/vllm/pull/727. This is the companion eligibility contract for #353, not a replacement for its kernels. @coderabbitai review

voipmonitor:
The serving evidence is now committed at docs/nvfp4-immutable-input-serving.md, with raw measured/warmup latencies, exact source and image boundaries, physical GPU UUIDs and +6000/600 W settings, launchers/clients, ratio direction, and the strict cross-kernel cosine limitation. The gains require #353 plus vLLM #727; no decode gain or bit-exact distribution claim is made. @coderabbitai review

voipmonitor:
Review disposition and minimal witnesses:

The launch eligibility check is **prepared-weight metadata**, not a GPU-component policy lookup or repeated scale validation. `_uniform_a1_scale` is computed once in the owner constructor. `can_share_input()` reads only tensor extent, stored booleans and (for versioned tensors) `_version`. It performs no `.item()`, reduction, allocation, synchronization or GPU value read. CUDA graph replay executes no Python predicate at all.

`test_binding_decision_does_not_read_scale_values` replaces `Tensor.item` and `torch.isfinite` with functions that raise; launch eligibility still works. `test_scale_mutation_invalidates_proof_until_weights_are_reprepared` supplies the mutation witness: a versioned owner's scale vector changes, eligibility becomes false, and merely restoring equal values does not restore the proof until weight preparation is repeated. The 39 installed CPU/GPU contracts include actual graph replay and allocation checks.

Caching another boolean in a reusable binding must not drop that invalidation check. vLLM also constructs caller-scratch bindings per call, so moving the same metadata read into its binding construction does not remove the work. The implementation therefore retains the fail-closed metadata guard; it does not add a replay-time policy query. Inference-tensor callers explicitly promise immutable values for the complete graph lifetime.

The documentation now defines concurrency-one (C1) and concurrency-eight (C8) before use and includes Qwen decode receipts. Evidence-client fixes reject missing cache evidence and invalid durations, guarantee a measured sample after warmup/receipt overhead, and validate TP1/TP2 devices in the correct branch. Six focused client tests pass. The copied GLM dependency was narrowed to its two prompt-counter functions, so an unrelated external-cache restore CLI is not shipped as B12X tooling. These are evidence-client/docs changes only: the measured serving code and release image are unchanged.

The warmup-expiration explanation needs one correction: the timer starts **after** the warmup completes. The added boundary test covers delayed warmup-receipt handling, not a warmup charged to the measured interval. All accepted 30-second measurements contain explicit zero-cache evidence and remain valid.

@coderabbitai review

PR #361, opened 2026-09-12
fix(gemm): support K32-aligned block-FP8 TP shards

voipmonitor:
## Behavior

Block-FP8 linear projections accept checkpoint shards whose input width is a
multiple of their 32-column scale block, even when it is not a multiple of the
128-column MXFP8 GEMM tile. The logical input width and tensor-parallel shard
offsets remain unchanged.

DeepSeek V4.1 Flash has a 2,304-wide shared expert. At TP4 its row-parallel
down projection has K=576. B12X master `081b2359` rejects that valid checkpoint
shape after loading the model because its block-FP8 packer requires K128.
The standalone TP loading regression also exposes the same defect at K=32.

## Implementation and compatibility

- Pad packed weights to K640 for the K576 shard, preserving serialized FP8
  values and UE8M0 scales. Additional weight columns are zero with unit scales.
- Quantize the logical activation directly into physically padded caller-owned
  scratch. The existing CuTe quantization launch writes the zero tail and
  finite scales; there is no separate runtime padding copy or fill.
- Keep the aligned K128 path mask-free through compile-time specialization.
- Size and validate the scratch and GEMM operands by physical width, while
  public input validation and policy queries retain the logical model width.
- Preserve the FP8 activation-scale floor, exact UE8M0 requirement, functional
  compile path, and caller-owned graph replay contract. No backend fallback or
  vLLM tensor-parallel loader change is introduced.

## Validation

Qualified on RTX PRO 6000 Workstation (SM120), CUDA 13.3 / PyTorch 2.13:

- 105 B12X tests pass: block-FP8 GEMM, scratch/capacity/policy, quantization
  parity, and DeepSeek V4.1 experts.
- 11 vLLM tests pass at JJ `377c93be`: TP4 shard loading and graph replay,
  the independent MoE oracle, mapped-RAM Engram aliases, capacity accounting,
  and checkpoint descriptor loading.
- Added independent K32-scale tests for K32 and K576, including
  `[4096,576] @ [5120,576].T`, BF16/FP16 poisoned-scratch graph replay, and
  multiple live row counts with kernel resolution frozen after warmup.
- Ruff and `git diff --check` pass.

The isolated vLLM TP fixture uses native NCCL because it captures a raw Torch
graph without a model-owned B12X collective pool. The separate serving test
below qualifies B12X PCIe graph capture. GB10 is not qualified.

## Serving qualification

The source-locked two-layer image contains unmodified vLLM JJ `377c93be`
and B12X `fcecbfe6` (master `081b2359` plus this PR), without source mounts.
All tracked files and modes match the Git trees. Local image ID:
`sha256:da403ee8ceff7b510aeae1a04ca751478b2413099cfde4b45b99b96397ea0bd4`.

Native DeepSeek V4.1 Flash starts at TP4/DCP1 with DSpark K7, 188.833 GiB
mapped pinned host-RAM Engram tables, 4,096 scheduler tokens, OMP8,
NCCL16/2 MiB, and B12X attention/MoE/linear/PCIe collectives. Target
PIECEWISE/FULL and DSpark FULL graph captures succeed. Deterministic arithmetic
and pigeon-image smoke requests pass, and the service stays healthy.

Two warmed C1/context-zero runs (15-second warmup, 30-second cells) on the
same RTX PRO 6000 quartet at +6000 VRAM, temperature 1/top-p 0.95:

| Measurement | Run 1 | Run 2 |
|---|---:|---:|
| Output tok/s | 237.57 | 240.98 |
| Verifier steps/s | 92.04 | 90.89 |
| Emitted tokens per step | 2.581 | 2.652 |
| Uncached 32K prefill tok/s, median | 12,836.71 | 13,332.42 |

Each prefill repetition includes an excluded warmup and 30 measured seconds
(12 and 13 requests respectively). Every request computes 32,768 tokens,
with zero cache reuse or external restore. No request errors or recorded GPU
throttle masks occur. These numbers qualify the complete serving composition;
they are not an isolated speedup claim for padding, since master alone cannot
start this TP4 model shape.

### DIFF
--- a/b12x/_lib/quant/mxfp8_rows.py
+++ b/b12x/_lib/quant/mxfp8_rows.py
@@ -43,8 +43,10 @@ class _MXFP8RowsQuantLaunch:
         threads: int,
         trellis_native_mma_order: bool,
         min_amax: float,
+        source_k: int,
     ) -> None:
         self._k = int(k)
+        self._source_k = int(source_k)
         self._groups_k = self._k // 32
         self._source_type = source_type
         self._subgroup_width = int(subgroup_width)
@@ -66,7 +68,7 @@ class _MXFP8RowsQuantLaunch:
     ) -> None:
         source = cute.make_tensor(
             source_ptr,
-            cute.make_ordered_layout((m, self._k), order=(1, 0)),
+            cute.make_ordered_layout((m, self._source_k), order=(1, 0)),
         )
         values_u32 = cute.make_tensor(
             values_ptr,
@@ -116,7 +118,7 @@ class _MXFP8RowsQuantLaunch:
                     values = cute.make_rmem_tensor((8,), cutlass.Float32)
                     k0 = group * Int32(32) + lane8 * Int32(8)
                     for elem in cutlass.range_constexpr(8):
-                        values[elem] = cutlass.Float32(source[row, k0 + Int32(elem)])
+                        values[elem] = self._load(source, row, k0 + Int32(elem))
 
                     max_abs = fabs_f32(values[0])
                     for elem in cutlass.range_constexpr(1, 8):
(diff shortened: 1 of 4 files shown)


voipmonitor:
@coderabbitai review

PR #363, opened 2026-09-13
fix(moe): restore small-tile W4A8 cooperative launch occupancy

eugr:
DeepSeek V4 startup fails on GB10 with CUDA_ERROR_COOPERATIVE_LAUNCH_TOO_LARGE.
Small-tile kernels retain branches referencing unused barriers 11 and 12, increasing the compiled barrier count from 11 to 13. This reduces residency from two blocks per SM to one, making the existing 96-block cooperative launch exceed capacity.
Specialize input-pair synchronization at compile time so small tiles exclude those unused barriers.
Validation: the isolated reproducer fails before this change and passes afterward; compiled barrier count returns to 11 and occupancy to two blocks per SM. Tested on the device, and was able to start the model.

## Summary

Specializes `_sync_input_warp_pair` synchronization barriers at compile time. Configurations with up to six route warps now use barriers 0–2. Larger configurations continue to use barriers 0–4.

This removes unused barriers from small-tile W4A8 kernels. The compiled barrier count returns from 13 to 11, which restores two-block-per-SM occupancy and prevents `CUDA_ERROR_COOPERATIVE_LAUNCH_TOO_LARGE` on GB10.

The isolated reproducer passes, and DeepSeek V4 starts successfully on the affected device. No exported or public declarations change.


eugr:
@lukealonso - can you please review? fixes the issue with deepseek v4 flash on Spark.

voipmonitor:
Independent RTX PRO 6000 Blackwell Workstation validation of **9dc276f8**: **no blocking implementation findings; I recommend keeping this fix.**

This fixes the same DS4 TP2 failure we reproduced in R35/R36, not just a GB10-specific issue. On one 188-SM RTX PRO 6000, using the R36 CUDA/PyTorch/CUTLASS runtime and the unmodified PR source:

| DS4 TP2 MoE case | Block barriers | Resident blocks/SM | Cooperative grid | Result |
|---|---:|---:|---:|---|
| R36 B12X, without the fix | 13 | 1 | 376 CTAs | CUDA error 720 |
| PR head | 11 | 2 | 376 CTAs | Pass |

Geometry: 256 experts, hidden width 4096, per-rank intermediate width 1024, top-6 routing, 16 input tokens, tile M32, 192 threads/block. No grid cap or diagnostic method replacement was applied to the PR. Registers remain 168/thread. Its complete generated CUBIN is **byte-identical to the passing pre-regression kernel at a3564a45**.

Also passed **21 focused tests**: 17 planner/boundary cases plus M64/M128 numerical-oracle checks, CUDA graph replay with mutated inputs/routing, and graph replay alongside auxiliary-stream GEMM work.

The predicate is correct: both call sites are in the W4A8 shared-input path, which uses two warps per input. Six route warps therefore mean three pairs. Larger configurations retain all five pair barriers. The selection is compile-time, with no per-request overhead. Deriving the condition from the pair count would be a maintainability alternative, not a required correctness or performance improvement.

Two small additions would make the patch more durable:

- An invariant comment explaining the two-warps-per-input assumption and why unreachable named barriers must be excluded.
- A GPU regression test for the exact DS4 TP2 M32 failure geometry, including uncapped cooperative launch and graph replay. The complementary tests above do not replace that specific regression test.

**Scope:** component correctness and launch validity, not an end-to-end speedup claim. We did not independently repeat full-model text/Vision startup or GB10 qualification in this review.

<details>
<summary>Reproduction and test selection</summary>

Runtime: `localinferencelab/vllm:jovian-judgement-community-20260912-r36`, with a clean checkout of PR commit `9dc276f8105cfbe2d5882a6475e6e96c9533911c` mounted read-only at `/opt/glm53-flash/b12x`. GPU: SM120, 188 SMs, driver 610.57.04. No clock changes; no throughput measurement is claimed. B12X disk compilation cache disabled, isolated CuTe cache, `NCCL_GRAPH_FILE` unset.

The launch probe wraps the following benchmark only to inspect the compiled kernel's resources; it forwards the selected kernel and grid unchanged:

```bash
python benchmarks/benchmark_ds4_moe.py --modes w4a8_mx --m 16 --warmup 1 --iters 1
```

Focused suite:

```bash
python -m pytest -p no:cacheprovider -vv -s \
  tests/moe/test_w4a8_mx_tp_moe.py \
  -k 'dynamic_tile_density_boundaries or ds4_tp2 or dynamic_graph_replay_tracks_routing_updates or m9_graph_replay_with_aux_stream_work or materialized_dense_override_matches_oracle'
```

Result: `21 passed, 39 deselected in 14.32s`. `git diff 9dc276f^ 9dc276f --check` also passes. CUBIN SHA-256 shared by the PR and the passing pre-regression control: `143559434193177eb90a6d0b0baa9049443e985f7182e1a3e30fcd008b0a2839`.

</details>

PR #366, opened 2026-09-13
perf(indexer): score only visible MXFP4 tiles

voipmonitor:
## Behavior
Bound MXFP4 sparse-index scoring to visible tiles and use the 32-head MMA geometry when supported. Keep valid scores, runtime lengths, top-k selection and caller scratch semantics unchanged. Invisible private score tails are not initialized output and must not be included in numerical comparisons.

## Validation
`python -m pytest tests/attention/test_dsa_indexer_mxfp4.py`: 55 GPU cases passed on RTX PRO 6000 Workstation. Coverage includes numerical scoring, masking, public dispatch and graph replay. Serving composition checks passed selector values, membership, counts, range and uniqueness on all four TP ranks, including live-address changes.

The focused score kernel was approximately 2× faster for the measured geometry. No isolated end-to-end serving gain is claimed. The integrated DSpark K7/TP4/DCP1/FP8-KV service passed four changing approximately 29K-token lookup requests. No precision conversion is introduced.

Source delta remains identical when composed over B12X `05d2c43b`, including its unrelated MoE barrier fix. No duplicate open DS4.1 indexer PR found. AI assistance: OpenAI Codex, under Martin Vit's requested component review.


voipmonitor:
@coderabbitai review

voipmonitor:
R37 integration qualification is complete. The tested two-layer image is
`localinferencelab/vllm:jovian-judgement-community-20260913-r37`, with committed
vLLM `c687594b8a8` and B12X `f1c4e9dd5b1`. Installed source trees match the
manifest; fourteen native binaries are unchanged from R36.

TP4/DCP1, RTX PRO 6000 Workstation VRAM +6000, DSpark K7/adaptive verification,
RAM Engram, batch 4096, temperature 1/top-p .95:

- Uncached 32K prefill: 21,399 tok/s; repeat after decode: 21,452 tok/s.
- C1 at numeric reasoning budget 50: median 281.08 output tok/s and
  110.80 verifier steps/s across three warmed 30-second cells.
- Sieve at budget 50: five-run median 416.72 tok/s (385.40–439.01).
- Corrected default budget 75 is separate: C1 258.79 tok/s / 104.11 steps/s;
  Sieve 393.58 tok/s (371.94–442.91).
- Text, image and seeded-sampling smoke passed; no measured engine errors.

This qualifies the **combined composition**, not an isolated gain attributable
to this PR. Default prefill graphs are off; target/draft decode graphs remain
enabled. Optional prefill-capture evidence remains separately documented in
vLLM #742. Generated Sieve code was not executed; other models, SSD throughput
and long-context LMCache were not requalified by these measurements.

Source disposition, comparison conditions and measurements:
https://github.com/local-inference-lab/vllm/issues/745

Artifact and raw measurement receipts:
https://github.com/local-inference-lab/rtx6kpro/tree/master/models/deepseek-v4.1-flash/r37

PR #372, opened 2026-09-14
ci: notify container resolver after wheel publication

voipmonitor:
## Purpose

Trigger the wheel-built Docker resolver immediately after a complete component release, instead of waiting only for GitHub's scheduled scans.

## Behavior

The notification job depends on successful publication, including verified reuse of an existing release, and only runs on the tracked source branch. It calls the commit-pinned reusable workflow from [blackwell-llm-docker#37](https://github.com/local-inference-lab/blackwell-llm-docker/pull/37). The credential is consumed by a separate GitHub-hosted job without checkout; the target repository, workflow and main ref are fixed.

Compiler resources, native caches, source locks, package contents and serving processes are unchanged. Stable promotions do not dispatch beta builds.

Status: **implemented; qualified by workflow lint and offline dispatcher tests**. Configure `LIL_CONTAINER_DISPATCH_TOKEN` with Actions write permission only in `local-inference-lab/blackwell-llm-docker` before enabling the notification. Live dispatch is not yet qualified.

## Validation

- Actionlint passes with the declared self-hosted runner labels.
- Applicable changed-file pre-commit hooks pass.
- Shared dispatcher executes four cases: success, missing token, transient retry and exhausted retries; the full runtime-tool suite reports **68 passed**.
- No model execution changes; model evaluation is not applicable.

AI-assisted implementation requested by the maintainer. No open PR addressing component-publication notification was found.

Adds a `notify-container` job that runs after successful beta publication on `master`. It dispatches the commit-pinned container resolver workflow with `LIL_CONTAINER_DISPATCH_TOKEN`.

The job has no permissions and does not run for stable-release tags. Existing build and promotion behavior remains unchanged. Validation includes actionlint, applicable pre-commit hooks, and dispatcher tests; live dispatch is not yet qualified.


voipmonitor:
Superseded by #381, which preserves this notification commit and adds integration/beta to the shared wheel publisher. Restricted cross-repository notification is explicitly enabled after its secret is configured. No kernel code changes.

PR #373, opened 2026-09-14
fix(comm): prepare two-shot executables and native row geometry

voipmonitor:
## Behavior and reason

Prepare BF16 and FP8 PCIe two-shot launchers with explicit executable ownership:

- Attach the compiled kernel to each host launcher through `attach_programs`.
- Load all eager and graph-slot variants during materialization, before kernel resolution is frozen.
- Prime BF16 all-reduce using native row views, matching its runtime path, and reject declarations with incomplete rank shards.

Without the dependency metadata, DS4.1 TP4 startup fails with `compile factory returned an unannotated function`. Adding metadata alone exposes another failure: a graph-slot executable is first loaded during CUDA capture. A separate GPU reproducer also found that priming a valid `16 x 5120` activation used token rows instead of all twenty native 4096-element rows.

The runtime kernel arithmetic, dtype, grid policy, precision, and caller-owned output contract are unchanged. The dependency/loading mechanism follows the existing oneshot implementation. Companion [vLLM #759](https://github.com/local-inference-lab/vllm/pull/759) makes metadata routing reject native rows that do not divide TP, as its runtime already does.

## Validation

Status: component-qualified on four RTX PRO 6000 Blackwell Workstation GPUs, stock clocks; complete DS4.1 startup remains pending.

```bash
python -m pytest -q tests/comm/test_twoshot_program_dependencies.py \
  tests/preparation/test_compile_admission.py \
  tests/preparation/test_first_use_compile.py
python -m torch.distributed.run --standalone --nproc-per-node=4 \
  tests/comm/test_prepared_twoshot_bf16_gpu.py
```

- Metadata discovery reproduces the original TypeError on master `9e90d60f` and passes with the fix: 40 transport/world/rank combinations, 120 eager/graph-slot variants, no CUDA execution or kernel lowering.
- Focused host suites: **11 passed**.
- TP4 BF16 all-reduce, reduce-scatter and all-gather: preparation, eager calls and three mutated-input CUDA graph replays pass. Reduction output satisfies the FP32-sum/BF16-rounding oracle; gather is exact. Caller output addresses remain stable and replay adds no PyTorch allocations.
- A priming check fails before the native-row reshape and passes afterward for `16 x 5120`.
- Existing `test_pcie_twoshot_unit.py` and `test_pcie_twoshot_bf16.py` have five legacy-API unit failures on both master and the patched source. Their torchrun harness also calls removed `prepare_graph()`. Those tests are not reported as passing; the added GPU test uses the prepared-plan API.
- FP8 transport has metadata-discovery coverage, not a fresh distributed numerical qualification in this PR.
- Ruff checks for the added tests and `git diff --check` pass.

## Cache compatibility

B12X selection records and CuTe cache identities include the package content fingerprint. This source change invalidates that identity; no cache fingerprint or validity check is bypassed, and no cross-revision reuse is promised.

AI-assisted implementation and testing. Existing collective implementation authorship is preserved.


voipmonitor:
TP4 DS4.1 startup qualification passed on four RTX PRO 6000 Workstation GPUs.

The tested composition contains B12X #373 and vLLM #759/#760. Both prepared
two-shot capture and the first Engram forward now complete. The server uses
DSpark K7, RAM-resident Engram tables, FP8 KV, DCP1, and a 4096-token scheduler
budget. No serving-source bind mounts or kernel/precision fallbacks were used.

- 19 piecewise target graphs, 33 full target graphs, 16 DSpark graphs and
  8 DSpark context graphs captured on all four ranks.
- HTTP startup, a deterministic `2 + 2` request returning `4` with
  `finish_reason=stop`, and post-request health all passed.
- All 1869 completed weights-stage selections per rank were reused without
  candidate measurements or compilations; this preparation took 35 seconds.
  First-time production-pool attention preparation completed separately in
  78 seconds. The original selection records compare equal to their backup;
  384 attention records per rank were added.

Tested vLLM: `465cf7508264cabadb0870c7ba393cff218448f7`.
Tested B12X: `a78476896d4ced0b7753d7c8db2855d61b58c4fd`.
Image ID:
`sha256:7766f2e4d9995cab9889e6180aae3afacdc1f8271bc099e09c9578bfa614ea02`.

This is composed startup and API smoke evidence, not an isolated performance
A/B, model-quality evaluation, or long-context qualification. AI-assisted
validation; no human manual-review claim.

voipmonitor:
Additional validation found a separate shape-contract issue at the prepared execution boundary. #380 fixes native transport row counting and includes a four-rank reproducer. All four producer shapes pass eager and three mutated CUDA-graph replays on RTX PRO 6000 Workstation. This complements the dependency/priming changes here; it does not supersede this PR or enable two-shot by default.

PR #380, opened 2026-09-16
fix(comm): count native rows in prepared BF16 two-shot

voipmonitor:
## Behavior

Count BF16 two-shot transport rows as `payload.numel() // runtime.row_elems` in the prepared launcher. Contiguous tensors with the same element count can have different producer shapes; their leading dimension is not necessarily the transport-row count.

Using `shape[0]` can launch too few rows or exceed the configured transport capacity when preparation invokes the state directly. Public all-reduce already reshapes its input, but prepared execution must obey the same shape-invariant contract without depending on that wrapper.

The reduction arithmetic, launch policy and communication thresholds are unchanged. This does not enable two-shot by default.

## Validation

Status: implemented and component-qualified on four RTX PRO 6000 Workstation GPUs.

```sh
python -m pytest --noconftest -q tests/comm/test_pcie_twoshot_bf16.py \
  -k prepared_all_reduce_counts_native_rows
PYTHONPATH=. torchrun --standalone --nproc-per-node=4 \
  tests/comm/validate_prepared_twoshot_native_rows.py
```

- Four CPU cases pass: `(128,1024)`, `(32,4096)`, `(16,8192)` and a flat 131072-element tensor all represent 32 native rows at width 4096.
- All four shapes pass direct prepared execution and three CUDA-graph replays with mutated inputs. Output is poisoned with NaNs before replay, and every element must equal the exactly representable rank sum: 10, then 14/18/22. The test script is included so this evidence is reproducible.
- A composed TP4 GLM server with a 768-KB two-shot threshold previously passed startup, graphs, text, vision, Sieve and prefill. It showed no material C1 benefit (89.14 versus 89.01 request-verifier steps/s); two-shot remains disabled in the served configuration. No performance gain is claimed for this correctness fix.

## Related change

This complements #373. That PR deals with compiled dependencies and preparation priming; this PR makes the prepared execution boundary itself independent of producer shape. Neither change is a substitute for the other.

AI-assisted implementation and validation; existing contributor ownership is retained.


voipmonitor:
@coderabbitai review

PR #382, opened 2026-09-16
experiment(dsa_indexer): allow 1024/2048 in the MXFP4 Caps and query gates

fernandaspets:
## What this is

Relaxes the two MXFP4 DSA-indexer **qualification gates** so the indexer top-k width can be
512, 1024, or 2048 instead of 512-only. **512 stays the only qualified width; 1024/2048 are
experimental.**

The gate is a *qualification boundary*, not a kernel constraint: `msa_topk_blocks` takes a
dynamic topk, the gather paths use `min(topk, width)`, and the persistent top-k primitives
are built for 2048. So 1024/2048 are accepted but unqualified.

Supersedes closed #376 (rebased onto current `master`; same content). Counterpart **vllm** PR: local-inference-lab/vllm#776 (the `index_topk` threading — **both are required** (see below).

## Changes (1 commit)

- `b12x/attention/dsa_indexer/api.py` — `Caps` guard `topk != 512` → `topk not in (512, 1024, 2048)`.
- `b12x/attention/dsa_indexer/_tuning.py` — `_validate_query` guard, same relaxation. Error
  messages name 512 as qualified and 1024/2048 as experimental.
- `tests/attention/test_attention_dsa_indexer_api.py` — 8 parametrized tests: `Caps` and
  `DsaIndexerQuery`/`_validate_query` accept 512/1024/2048 for `cache_format="mxfp4"` and
  still reject 1/256/768/4096.

## ⚠️ BOTH repos are required

b12x accepting the wider value is necessary but **not sufficient**. The vllm side must thread
`index_topk` into the indexer `Caps`, the MLA plan widths, **and the preparation primers** —
two primer sites hardcode 512 and will kill the boot at 1024 even with these gates relaxed
(`mxfp4_index.decode.m1 … output_indices must have shape (1, 1024)` and
`mla.decode.m240 … runtime unified decode metadata differs from its prepared launch`).
The vllm PR carries those fixes.

## How to use it

With both branches built in, set the width at serve time with a **flat** `--hf-overrides`
key (vllm's `DeepseekV41Config` flattens `text_config`, so a flat key is what the model reads):

```bash
--hf-overrides '{"index_topk": 1024}'    # boot-tested
--hf-overrides '{"index_topk": 2048}'    # same mechanism, not yet boot-tested
```

**Precedence:** `--hf-overrides` > `config.json` `text_config.index_topk` > code default 512.
Confirm with `VLLM_LOGGING_LEVEL=DEBUG` → the engine logs `Overriding HF config with {'index_topk': N}`.

## Tested

- This branch on current `master`: `tests/attention/test_attention_dsa_indexer_api.py`
  **27 passed, 27 skipped**.
- With the vllm branch at **1024**: engine boots, effective value 1024, KV pool unchanged
  (6,633,904 tokens), correct arithmetic smoke, canonical bench C1 median 123.5 tok/s,
  prefill ladder 8k 10,241 → 128k 8,810 tok/s.

512 remains the only qualified width; 1024/2048 are experimental and the per-width quality
suite should be re-run before either is adopted.

## Summary

MXFP4 DSA-indexer validation now accepts `topk` values 512, 1024, and 2048 when `output_index_space` is logical and `page_size` is positive. Width 512 remains qualified. Widths 1024 and 2048 remain experimental and unqualified. Other widths remain rejected.

The change updates `Caps.__post_init__` and `_validate_query`. It changes only the qualification boundary. Existing capacity policies, persistent top-k support, and kernel behavior remain unchanged. Wider widths use generic execution paths where specialized 512 kernels do not apply.

Tests cover accepted widths 512, 1024, and 2048, plus rejected widths 1, 256, 768, and 4096, through both validation paths. Reported validation includes 27 passed and 27 skipped API tests and a width-1024 boot test. Per-width quality testing is still required before 1024 or 2048 can become qualified or default widths.


fernandaspets:
Superseded by #385, rebased on current `master` with the review feedback addressed: the query gate now rejects physical output slots and the tests cover it. Closing in favor of the clean branch.

PR #391, opened 2026-09-17
feat(gemm): capacity regimes for fixed blockscaled plans

tobymao:
**TL;DR:** `plan_regimes` now accepts a `FixedBlockscaledQuery` with `expected_m` unset. That lets a caller declare one fixed plan per layer: exact static-shape plans for its fixed row counts, plus one capacity plan that runs any row count up to `max_rows`. Until now a fixed plan covered exactly one row count. A vLLM integration serving arbitrary prefill lengths therefore prepared a new plan on the request path for every length it hadn't seen. On a 4x GB10 GLM-5.3-Flash native FP8 deployment that cost about 0.66 s per new prefill length. With this change and the vLLM caller, the cost is gone.

## Behavior

- **`plan_regimes(query, exact_m=...)`** accepts `BlockscaledQuery` as before, and now also `FixedBlockscaledQuery`. The capacity query must leave `expected_m` unset. Each count in `exact_m` gets its own static-shape child. Every other row count up to the capacity runs the capacity child. The composite uses `_CompositePlan` with `capacity_metadata`, like the existing BF16 regime.
- **Resolving a call:** the fixed custom ops (`blockscaled_serialized`, `packed_mxfp8*`, `packed_tensor_fp8`) pick the exact child for the live row count if there is one, otherwise the capacity child. Above the capacity they raise `rows N exceed capacity C`. A plain, non-regime fixed plan behaves exactly as before.
- **Which recipes accept a capacity query:**
  - Serialized `block_fp8`, `nvfp4`, `mxfp4`: the caller passes per-call scale storage for the live rows, and the dense program already accepts any `m <= p.m`.
  - Packed `tensor_fp8`: its unit activation scale is sized for `max_rows`. Row block `j` sits at the same offset whatever the row count, and the kernel reads only the live rows.
  - Packed MXFP8 keeps exact row counts only. Its FP16 quantizer is compiled per row count, and BF16 already has the `BlockscaledQuery` regime.
- **Validation:** `_validate_fixed_query` allows `expected_m=None` only for those recipes. `_FixedExecutionState._check` accepts any row count up to `max_rows` for a capacity state and keeps the exact-count check otherwise.
- **Unchanged:** arithmetic, kernel selection and existing exact declarations. `tensor_fp8_linear` re-exports `plan_regimes`. Unused imports in the blockscaled linear and tuning modules are removed.

## Why

A fixed plan's `expected_m` is a compile-time specialization. vLLM's block-FP8 and tensor-FP8 linears therefore declared one plan per row count and prepared it on first use (`gemm.blockscaled.fixed: FixedBlockscaledQuery (max_rows=N, ...) was not prepared before its first use; preparing its default configuration`). Decode and graph-capture sizes are prepared at startup, but a prefill chunk's length is effectively arbitrary. NVFP4 and MXFP8 BF16 linears avoid this with the `BlockscaledQuery` capacity regime; the fixed recipes had no equivalent.

## Validation

**Unit tests, GB10 / SM121 (4x DGX Spark), CUTLASS DSL 4.7.0, torch 2.13.0:**
- `tests/gemm/test_blockscaled.py tests/gemm/test_tensor_fp8_linear.py -k "regime or capacity"`: **13 passed, 0 skipped**, run on the GPU.
  - Serialized block-FP8 and packed tensor-FP8 capacity plans run M = 1, 3, 4, 17, 129, 257, 300 and match exact-M results and a reference under frozen kernel resolution.
  - M above capacity raises.
  - Plan shape for all four recipes, invalid declarations, exact-or-capacity resolution, row checks per child, op-body resolution, and tensor-FP8 memory accounting per row bound.
- `tests/gemm` + `tests/preparation` (CPU): 18 failed / 859 passed / 539 skipped on master, 18 / 868 / 541 here. The failures are the same set before and after.

**Serving, GLM-5.3-Flash native FP8** (TP4, MTP 3, 4x GB10, vLLM local-inference-lab `dev/jovian-judgement` 8e1f1e587f plus the caller change):
- Before, on the same stack with only exact-count plans:
  - 150 uncached prompts with 150 new prefill lengths: **332 s**. The identical-length repeat: **233 s**. The difference is about 0.66 s per new length.
  - 596 `was not prepared` lines (149 new lengths x 4 projections per layer type).
- After, the same prompts:
  - **226 s** first pass (includes post-boot warm-up), **177 s** repeat.
  - Another 150 lengths never seen before: **168 s**, level with the repeat.
  - **0** `was not prepared` lines across 500+ requests.
- Startup, CUDA graph capture, corruption probe and serving smoke pass.

## Notes

- The capacity child picks its tile and policy for `max_rows`, not the live chunk size. Measured prefill above is not slower (the repeat pass went from 233 s to 177 s).
- The vLLM caller change follows; it depends on this PR.

AI assistance was used in preparing this contribution.

## Summary

- Adds capacity regimes for fixed blockscaled GEMM plans.
- `FixedBlockscaledQuery.expected_m` can be unset. Such plans serve row counts from 1 through `max_rows`.
- Fixed custom operations select an exact child when available. Otherwise, they select the capacity child. Requests above `max_rows` fail.
- Capacity plans support serialized block-FP8, NVFP4, MXFP4, and packed tensor-FP8. Packed MXFP8 remains exact-count only.
- Execution-state validation now distinguishes exact plans from capacity plans.
- `tensor_fp8_linear` now exports `plan_regimes`.
- The change avoids request-path plan preparation for unseen prefill lengths in the supported vLLM integration.

## Compatibility

Existing exact row-count declarations, arithmetic, and kernel selection remain unchanged. Capacity queries require serialized operands or tensor-FP8.

## Validation

Tests cover query validation, regime planning and resolution, capacity enforcement, unit-scale sizing, frozen resolution, and serialized block-FP8 execution.


tobymao:
The vLLM caller is local-inference-lab/vllm#788. It moves block-FP8, tensor-FP8, serialized NVFP4 and MXFP4 linears onto this regime. With both applied, the GLM-5.3-Flash FP8 deployment logs 0 `was not prepared` lines, and its GPU tests for NVFP4/MXFP4 capacity rows pass (15/15).

PR #397, opened 2026-09-19
perf(moe): tune repacked W4A8 grids and preserve prepared capacity

voipmonitor:
## Tune physical CTA counts for repacked W4A8 decode

The prepared MoE tuner can now select the physical CUDA block count for
repacked W4A8-MX decode kernels with 16- or 32-row tiles. Previously, that
runtime knob was enumerated only for compact W4A8 or external Triton routing;
the normal repacked launch also ignored it. A DeepSeek V4 TP2 trace therefore
used 376 target blocks where the community implementation used 128.

The candidate set retains the uncapped configuration and adds a bounded
ladder through two blocks per SM. Selection stays in B12X preparation; replay
uses a scalar grid limit, with no additional compilation, allocation or
model-specific pin. Existing explicit grid environment overrides take
precedence. Prefill, deterministic output, ready-queue routing, compact W4A8
and NVFP4 retain their existing selection domains.

The GPU regression test also exposed a capacity-contract defect: lowering a
prepared multirow plan to one live row could change the intermediate
representation and request an uncompiled kernel. Representation selection
now uses the declared capacity, while live rows still control launch masks.
Exact one-row declarations retain their specialization. The capacity is
`routed_rows / top_k`, not the independent maximum chunk length.
Grid-override lookup follows that same prepared capacity: short tails of a
prefill-capacity plan do not switch to the decode-specific environment override.

This changes launch policy and preparation, not checkpoint format, activation
precision, router weights or sampling. The internal dynamic-launch operator
ABI changes together with its caller and fake implementation. The candidate
contract version invalidates stored choices without adding live values to
kernel keys.

## Validation

Status: **implemented**; 24 GPU numerical/replay cases and five-window DS4
serving behavior are **qualified**. The isolated grid change did not improve
verifier throughput in this serving configuration.

- Repacked N1024, K4096 tests cover capacities 1/8, live counts 1/2/6/8,
  M16/M32 tiles and physical grids uncapped/1/64/128. Preparation freezes
  before the first launch. All live counts reuse the same compiled callable.
- Three graph replays per count preserve pointers, allocate no CUDA storage,
  leave inactive output rows poisoned, and produce finite nonzero results
  with cosine similarity above 0.998 to the independent quantized reference.
- Eight compact-N64 cases retain their numerical and frozen-capacity behavior.
- CPU predicate coverage checks 48-SM and 188-SM devices, domain exclusions,
  grid bounds and the shared compile identity of runtime-grid candidates.
- The broader CPU preparation suite has 535 passes, 59 skips and one failure
  also reproduced on the base: the device-reclaim test expects a cached
  selection with autotuning disabled, while the implementation selects the
  default. It is not a fully passing suite.

DeepSeek V4 Flash, TP2, five DSpark drafts, RTX PRO 6000 Max-Q GPUs, VRAM
+6000, automatic graphics clocks, 325 W, temperature 1/top-p 1:

| Five-window median | Prepared WO storage repair | Plus grid tuning |
|---|---:|---:|
| Uncached 32K prefill, tokens/s | 11,362 | 11,395 |
| C1 output, tokens/s | 204.789 | 191.045 |
| C1 verifier, steps/s | 71.694 | 71.684 |
| C1 accepted length | 2.863 | 2.678 |
| C8 output, tokens/s | 655.399 | 669.416 |
| C8 verifier, steps/s | 247.623 | 246.848 |

All functional checks and timed cells pass. C1/C8 verifier rates change
-0.014%/-0.313%. Output differences coincide with different acceptance in
unseeded requests; these samples do not prove a distribution change. The
rank-zero trace confirms target grids changed from 376 to 188 blocks, but
that change is not itself a demonstrated regression fix.

Commands, individual serving samples, source/image identities and limits are
in the [committed qualification report](https://github.com/local-inference-lab/b12x/blob/perf/w4a8-decode-resident-grid/validation/serving/w4a8_resident_grid/README.md)
and its raw JSON. The isolated branch is based on master
`0f3a8cbfd1c11d27f04e3ab37a802d522f4f1c68`. The subsequent override-namespace
correction is host-tested separately; it does not alter the recorded
override-free serving grids or their arithmetic.

```bash
uv run python -m pytest tests/moe/test_w4a8_mx_tp_moe.py \
  -k 'repacked_decode_grid_reaches_launch or compact_n64_capacity_plan' -v
uv run python -m pytest tests/moe/test_fused_moe_variant_selection.py \
  tests/preparation/test_tuning_predicates.py -q
```

The first command passes 24 GPU cases. The second passes 155 host-only cases,
including five capacity/override namespace cases. Two short-tail cases fail
without the namespace fix. Python syntax checks pass. Explicit environment
overrides retain precedence within the prepared execution regime.
This is distinct from #282/#334, which expose W4A16 reservation or pipeline
knobs, and #256, which addresses routed-workspace indexing rather than this
prepared W4A8 representation and resident-grid contract.

AI-assisted implementation and validation requested by @voipmonitor. The
repository PR search found no duplicate of this repacked-decode grid change.

## Summary

Prepared repacked W4A8 MoE decode launches now select physical CUDA block counts for M16 and M32 tiles, up to two blocks per SM. Environment grid overrides remain supported.

Kernel preparation uses declared capacity, while launch masks use live row counts. This preserves compiled representations when live batches shrink. Grid-override lookup also uses prepared capacity.

The dynamic-launch ABI, caller, fake implementation, tuning contract version, and candidate validation were updated. No checkpoint, precision, router, sampling, compilation, or allocation behavior changed.

## Validation

- 24 GPU numerical and replay cases passed.
- Compact W4A8 and CPU predicate coverage passed.
- Combined variant and tuning-predicate coverage passed 155 tests.
- Five-window DeepSeek V4 serving validation completed.
- The broader CPU preparation suite retains one pre-existing failure.
- The tested serving configuration showed no verifier-throughput improvement from the grid change.


voipmonitor:
@coderabbitai review

voipmonitor:
Verified the capacity finding against the runtime path and fixed it in 8d831a6f. With capacity 16/top-k 6, live counts 1 and 8 incorrectly chose the decode override namespace; live 16 chose generic dynamic. The host launch reproducer fails those two cases before the change and passes all five cases afterward, including unchanged capacity-8 decode behavior. Combined variant/predicate coverage: 155 passed. No capacity clamp was added. The recorded override-free serving grids and arithmetic are unchanged; this is not a new GPU performance claim. The report now also labels its columns with the raw evidence keys. @coderabbitai review

voipmonitor:
Verified the documentation finding and added launch-grids.json plus its extractor in fe35923d. Both rank traces contain four complete steps: each arm has 172 target and 12 draft launches per rank. The target grid is 376 blocks in the baseline and 188 in the candidate; draft remains 240. The artifact records ordered launches, trace hashes, and matching image/source identities. No runtime code or throughput samples changed. @coderabbitai review

